PR #6, opened 2022-06-21
Add support for different Docker versions

qts8n:
## Change COPY line to achieve compatibility with different Docker versions.

Apparantly some different Docker versions do not support this format of destination directory, therefore `.` has been changed to `./`.

Otherwise it was resulting in the following error:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /home/node
 
 VOLUME /home/node/node_modules
 
-COPY package*.json .
+COPY package*.json ./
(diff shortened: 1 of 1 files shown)


Daniel31x13:
You're right thanks for the tip!

Daniel31x13:
Can you redo the same pull request? I made a mistake

qts8n:
> Can you redo the same pull request? I made a mistake and you didn't show up as a contributor.

Sure thing! Here you go #8

PR #10, opened 2022-06-26
Fix permissions error when running in docker

gardner:
I rebuilt from scratch to test this out on mac but I don't have a windows / linux machine to test it on.


Daniel31x13:
Oh right I've already fixed it and everything works perfectly.

gardner:
Right on

Issue #94, opened 2023-07-31
sqlite as db


Code-Slave:
**Is your feature request related to a problem? Please describe.**
for personal use does pg provide anything that sqlite cannot? I try and get sqlite support from every self hosted app I can as pg is overkill 99% of the time

**Describe the solution you'd like**
Support for sqllite

I currently use linkding in sqllite for my family and have no issues. this apps looks great and Id like to use it due to the archiving.

daniel31x13:
Swapping SQLite for Postgres shouldn't be much of a problem, I'll add this to the project to-do.

crkos:
Hello i'd like to work on this, but i saw that more support for pg is being done, adding support for SQLite will add some changes to the database and some logic in the application, so i don't know if you really want to change from postgresql or change to SQLite in the future, definitly it will reduce some complication with working with a external instance of a db, as to just working with a file

daniel31x13:
Linkwarden now supports SQLite, you just need to change [This Line](https://github.com/linkwarden/linkwarden/blob/89427f16f5752405ddcdbae7e147621dc9a3f1e0/prisma/schema.prisma#L6) from `postgresql` to `sqlite` and use the correct `DATABASE_URL` while taking the [Manual Installation](https://docs.linkwarden.app/getting-started/self-hosting#manual-installation).

Also there is one drawback from using SQLite, things are case-sensitive when you search for a Link.
The reason for this is Prisma (currently) doesn't support querying the database using the `mode: 'insensitive'` option for SQLite.

Gotta update the docs later...

Code-Slave:
awesome thanks

Code-Slave:
Ideally the db type would be an env in the compose. will make running with docker much harder

Mat-DB:
I came across this github repo and it looks really nice!
I just do not want to install manually and would like to use SQLite because postgresql is overkill for me.

So my question, is SQLite already supported using docker-compose?

Thanks!

lockheed:
Another thums up. Is the "env" route to SQLite possible yet?

akhy:
> Linkwarden now supports SQLite, you just need to change [This Line](https://github.com/linkwarden/linkwarden/blob/89427f16f5752405ddcdbae7e147621dc9a3f1e0/prisma/schema.prisma#L6) from `postgresql` to `sqlite` and use the correct `DATABASE_URL` while taking the [Manual Installation](https://docs.linkwarden.app/getting-started/self-hosting#manual-installation).
[...]

PR #115, opened 2023-08-04
add sqlite compatibility + fix whitespace bug collections

crkos:
I added sqlite compatibility, SQLite doesn't have compatibility for array of primitives so instead whitelistedNames had to be done in a separated table to be able to use them as an array of them.

Also added some files on gitignore, only model whitelistedUser was added to the DB nothing else was moved except some decorators of text typings because SQLite doesn't take them.

Finally a bug fixed causing to spin infinitely on the UI when you add a link and add whitespace or write on the search bar of collection, and send the link it will trigger a bug because whitespace or null was sent in the request body and will try to trim that and will throw an error

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,8 @@ next-env.d.ts
 
 # generated files and folders
 /data
+.idea
+prisma/dev.db
 
 # tests
 /tests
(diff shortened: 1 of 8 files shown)


crkos:
This PR address #94

daniel31x13 (approved):

PR #131, opened 2023-08-07
Feat: Dark mode

g1sbi:
I have implemented dark mode across the site, however, there are a few caveats:

- I haven't been able to stylize the `react-select` components, namely in `TagSelection.tsx` and `CollectionSelection.tsx`.
- When adding a new link, a custom function is used to create a colored background. I have not modified it, so the links' backgrounds are mainly white as a result.
- For some reason, I wasn't able to access the team management's UI in `TeamManagement.tsx`. I have stylized it, but there could be issues since I had to do it blindly.


daniel31x13:
Thank you!
Reviewing it…

PR #168, opened 2023-09-20
Use official nodejs base image, slim down container image

peschmae:
- Playwright 1.35 supports debian 11 (bullseye), for debian 12 (bookworm) support, playwright needs to be updated
- Ensure all apt & yarn caches are cleaned in the same step they are used (slims down image)
- Image uses a lot less steps now, and slimmed down form ~4GB to 2.5GB

This makes the container image easier to manage, as it has less steps, and updating just relies on updating the upstream images.

I'm already running this image, and it works fine including generating the screenshots & pdf for websites.


peschmae:
It also makes building the images faster and will help with #166

daniel31x13:
Merging it, thanks.

daniel31x13:
@peschmae Hey just tested this on an M1 Mac and noticed it crashes when I start archiving a webpage, do you know what could be causing the problem?

peschmae:
Oh, I only tested it on amd64 hardware, as I thought I read that arm64 isn't supported

But I'll give this a try tomorrow, as I have access to an M1 at work and check it out.

The previous image worked on M1, I assume?

daniel31x13:
Sorry for the late reply, looks like Playwright isn't working properly on macOS, which is unusual given that Docker should work consistently across all platforms. Just tested is and it looks like the previous image didn't capture the webpages as well.

peschmae:
There seem to be a few tickets for playwright and arm64/m1, for example: https://github.com/microsoft/playwright/issues/11150

I'll try switching playwright to firefox in the docker container, to see if it works then.

peschmae:
I've created a issue to follow up on this: #175

PR #236, opened 2023-10-19
html form for login & register using enter key

YeeJiaWei:
I have update the login and register page to use html form for login and register with ability to press enter to submit.
i have also add autofocus to first input of pages, which let user to type credentials on page loaded.


daniel31x13:
@YeeJiaWei Thanks! will be reviewing it soon.

PR #266, opened 2023-10-28
Added support for other database providers

robertripoll:
- Updated `.env.sample` to add a database provider variable
- Modified `Dockerfile` so the database provider is injected in Prisma's schema
- Modified Prisma's schema to add a placeholder where the database provider can be injected
- Removed Postgres dependency from `docker-compose`


robertripoll:
When running with a different database provider other than `postgresql`, Prisma throws this error:
```
The datasource provider `sqlite` specified in your schema does not match the one specified in the migration_lock.toml, `postgresql`. Please remove your current migration directory and start a new migration history with prisma migrate dev.
```

PR #275, opened 2023-11-07
feat: Basic support for Keycloak (OIDC) + fix s3 integration + custom s3 (minio) support

siccous:
Very simple OIDC implementation through Keycloak (I am not a JS guy so not sure if implemented right, but it works for me). There are lot of additional improvements which should be made (update User odel with data from Keycloak after every logon, disable editing username and password, etc...).

Edited few things around S3 integration to be able to use Minio instead of AWS S3.


daniel31x13:
Hey thanks for the contribution! Can you update the code as it now has conflicts with Linkwarden v2.0.0?
Also can you merge to the `dev` branch instead?

I'll be reviewing it soon.

PR #321, opened 2023-11-26
Added Authentik provider and option to disable standard login

Jacq:
Added Authentik provider
Added option to disable standard login with NEXT_PUBLIC_DISABLE_LOGIN=true


daniel31x13:
Hey thanks!
I'll be merging this alongside the other upcoming features in the next release.
Is there any way to contact you incase I had any questions?

Jacq:
I am available here, the changes are much of a copy from the keycloack merge, except I included also the new env var to disable standard email login.

Be ware that the chrome/firefox extensions required in the config a mandatory username/password but using the single sign on this should not be enforced, so currently this extensions are not compatible with SSO scenario without user/pass login.

PR #327, opened 2023-12-03
feat: refactored login

shxshxshxshx:
- Added all providers supported by NextAuth with the exception of those requiring custom callback handling (TODO tags in `enviornment.d.ts`).
- Users can define a custom name for SSO providers via the `<PROVIDER>_CUSTOM_NAME` environment variable.
- Account management on SSO-managed values (name/username/email/password/avatar) is no longer possible.
- Added the ability to deny SSO login for users that aren't already known.
- Added the ability to disable Credential authentication.
- Server-side rendering for parts of `pages/login.tsx` - NEXT_PUBLIC variables are resolved on `next build` and thus unavailable to the official docker image.


daniel31x13:
@shxshxshxshx thanks for the contribution!

I’ll be merging both PR in the next release along with the other upcoming features, because right now there are many features that we’re working on (like #318).

For now could you check if this has any conflicts with the PR in #321?

Thanks again!

daniel31x13:
@shxshxshxshx could you please resolve the conflicts and update the `.env.sample` file as well?
Thanks!

shxshxshxshx:
@daniel31x13 conflicts are resolved and `.env.sample` updated.

samip5:
Hi @shxshxshxshx, I tried out this already with a custom container, and it seems that next-auth at least will not like the redirect Authentik does.

Issuer as `https://auth.example.com/application/o/linkward/` will result in 301, and next-auth expects 200.


```
[next-auth][error][SIGNIN_OAUTH_ERROR]
https://next-auth.js.org/errors#signin_oauth_error expected 200 OK, got: 301 Moved Permanently {
  error: {
    message: 'expected 200 OK, got: 301 Moved Permanently',
    stack: 'OPError: expected 200 OK, got: 301 Moved Permanently\n' +
      '    at processResponse (/data/node_modules/openid-client/lib/helpers/process_response.js:41:11)\n' +
      '    at Issuer.discover (/data/node_modules/openid-client/lib/issuer.js:152:20)\n' +
      '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n' +
      '    at async openidClient (/data/node_modules/next-auth/core/lib/oauth/client.js:16:14)\n' +
      '    at async getAuthorizationUrl (/data/node_modules/next-auth/core/lib/oauth/authorization-url.js:70:18)\n' +
      '    at async Object.signin (/data/node_modules/next-auth/core/routes/signin.js:38:24)\n' +
      '    at async AuthHandler (/data/node_modules/next-auth/core/index.js:260:26)\n' +
      '    at async NextAuthApiHandler (/data/node_modules/next-auth/next/index.js:22:19)\n' +
      '    at async NextAuth._args$ (/data/node_modules/next-auth/next/index.js:106:14)',
    name: 'OPError'
  },
  providerId: 'authentik',
  message: 'expected 200 OK, got: 301 Moved Permanently'
}
```

shxshxshxshx:
@samip5 Nextauth expects the issuer without a trailing slash.

samip5:
> @samip5 Nextauth expects the issuer without a trailing slash.

Ah, yeah. That was the issue. Thanks for the pointer.

mpldr:
@shxshxshxshx is there a reasoning for adding all the providers instead of a single unified OIDC/OAuth2 provider?

daniel31x13:
@mpldr he just added every provider nextauth has, that's far better than just adding X, and then another user comes up and asks to add Y alongside X, and so on...

Basically more integrations is better.

mpldr:
The thing is that all of them are OAuth2/OIDC, just that with the specific providers some things are prefilled. That means all of them require basically the same information:
- issuer
- client ID
- client secret
- (optional) scope

Additional benefit to a generic interface: If there's a bug in one of the next-providers, users can use the generic implementation instead.

PR #339, opened 2023-12-11
fix: update manifest name for PWA display

treyg:
Just updated the `name` and `short_name` in the manifest file so it shows 'Linkwarden' when installed as a PWA. Right now, it just says 'website'


daniel31x13:
Hey there,
Sorry for the delay and thanks for pointing this out!

PR #344, opened 2023-12-16
Fix unable to save archives

YeeJiaWei:
this fix the archives are unable to save when directory not exists


daniel31x13:
Oh this is intentional, there is an upcoming feature that's gonna auto archives every link, even the ones that were imported.
I haven't stitched things together since this is the dev branch...

PR #390, opened 2024-01-01
fix: load environment variables in the worker script

QAComet:
Here's the changes to get the worker running correctly.


daniel31x13:
thanks for pointing this out!

PR #445, opened 2024-01-29
Added env var switch to support screen captures from HTTPS sites with untrusted certificates

jan-tee:
This is useful to support generation of browser screenshots from sources with self-signed certificates or untrusted CAs, but also opens the possibility to index sites with rejected certificates; so it should not be enabled as a default behavior.

To use, set `IGNORE_HTTPS_ERRORS` to `true`. Useful in self hosting, should never be enabled when running a site aimed at bookmarking internet sites.


daniel31x13:
Merging this, sorry for the delay.

PR #454, opened 2024-02-04
Add redirects to dashboard headers

### LINKED ISSUE #410: [Feature Request] Links in dashboard
Currently, on the dashboard we have:

image

Would it make sense from a QoL PoV to have the `Links`, `Collections` and maybe the `Tags` headers to be clickable, and go to `All Links`, `All Collections` etc?

IsaacWise06:
This closes #410, I didn't know what the tag header should link to so I left that empty.


daniel31x13:
> I didn't know what the tag header should link to so I left that empty

That's exactly why I haven't touched this yet :)

Issue #436, opened 2024-01-22
Edit Link itself & Collection folder color add RGB code


ouroboros-prime:
I really appreciate the efficiency of team version updates. When I started using 2.4.8 and found that I couldn't add local links directly, I wanted to change the software for a while, but the very fast update kept me here.

There are two functions that want to be enhanced here
1.Edit Link You cannot edit the link itself. When a link is invalid, you can only delete it and re-add the correct link. This should be directly editable.

2.Collection folder color. I would like to add RGB encoding to unify colors when classifying the same type.

daniel31x13:
Link editing is already implemented

PR #455, opened 2024-02-04
feat(links): Allow URLs to be edited

IsaacWise06:
Closes the first part of #436, I also changed the URL field to be required, I don't see/understand how it would've been able to be null/empty.

### DIFF
--- a/components/LinkViews/LinkCard.tsx
+++ b/components/LinkViews/LinkCard.tsx
@@ -21,9 +21,15 @@ type Props = {
   link: LinkIncludingShortenedCollectionAndTags;
   count: number;
   className?: string;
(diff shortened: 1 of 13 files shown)


daniel31x13:
Hey, first off thanks for putting your time into this!

URLs *should not* be able to be edited, the reason for that is that it's the main field of each bookmark and if you change it, the preserved formats and the URL will be out of sync...

And yes, we *can* change it so that a re-preservation happens, but the problem is when the users accidentally touch the URL and then see that they lost their preserved formats - this leads to a bad UX.

So yeah, the best way to "change the URL" is to just create a new link/bookmark.

PR #460, opened 2024-02-08
fix(collections): Fix prisma query for updating parent collections

IsaacWise06:


daniel31x13:
already fixed, thank you

PR #150, opened 2023-08-24
feat: proxy support for archiver

go-compile:
## Added Features
Added socks5 proxy support for pdf and png archiving (including the title fetch).
PDF margin ENV options.

### New environment variables
```ts
      ARCHIVER_PROXY?: string;
      ARCHIVER_PROXY_USERNAME?: string;
      ARCHIVER_PROXY_PASSWORD?: string;
      ARCHIVER_PROXY_BYPASS?: string;
      
      ARCHIVER_PDF_MARGIN_TOP?: string;
      ARCHIVER_PDF_MARGIN_BOTTOM?: string;
```

## Limitations
One limitation is the remote favicons being loaded via the gstatic API, this is not proxied.
A environment variable i.e. `DISABLED_REMOTE_FAVICON=true` would be the simplest fix for this, with the icons being replaced with a placeholder image instead.

PR #478, opened 2024-02-18
feat: Socks5 proxy support

go-compile:
Reopen of https://github.com/linkwarden/linkwarden/pull/150 on latest base.

## Added Features
Added socks5 proxy support for pdf and png archiving (including the title fetch).
PDF margin ENV options.

### New environment variables
```ts
      ARCHIVER_PROXY?: string;
      ARCHIVER_PROXY_USERNAME?: string;
      ARCHIVER_PROXY_PASSWORD?: string;
      ARCHIVER_PROXY_BYPASS?: string;
      
      ARCHIVER_PDF_MARGIN_TOP?: string;
      ARCHIVER_PDF_MARGIN_BOTTOM?: string;
```

## Limitations
One limitation is the remote favicons being loaded via the gstatic API, this is not proxied.
A environment variable i.e. `DISABLED_REMOTE_FAVICON=true` would be the simplest fix for this, with the icons being replaced with a placeholder image instead.


daniel31x13:
Thanks! can you just please replace npm with yarn?

go-compile:
Sure. I don't typically use yarn but I suppose you just need me to run `yarn install node-fetch socks-proxy-agent` and commit all changes?

daniel31x13:
The yarn.lock file is using package registries from npm instead of yarn so yeah you need to install the packages using yarn instead.

go-compile:
Done 👍

daniel31x13:
thank you!

PR #487, opened 2024-02-24
Archive as Epub

### LINKED ISSUE #496: Epub archive support
**Is your feature request related to a problem? Please describe.**
It'd be great to have the possibility to create epub archives of saved links. Epub is compatible with all e-book readers and is optimized for reading, it is the best format for storing news, blog posts, articles, etc.

**Describe the solution you'd like**
A similar solution to PDF archiving:
image

**Additional context**
Reference implementation: #487

rutkai:
Add the option to archive links as Epub books:
image

Fixes #496.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 
 **Linkwarden is a self-hosted, open-source collaborative bookmark manager to collect, organize and archive webpages.**
 
-The objective is to organize useful webpages and articles you find across the web in one place, and since useful webpages can go away (see the inevitability of [Link Rot](https://www.howtogeek.com/786227/what-is-link-rot-and-how-does-it-threaten-the-web/)), Linkwarden also saves a copy of each webpage as a Screenshot and PDF, ensuring accessibility even if the original content is no longer available.
+The objective is to organize useful webpages and articles you find across the web in one place, and since useful webpages can go away (see the inevitability of [Link Rot](https://www.howtogeek.com/786227/what-is-link-rot-and-how-does-it-threaten-the-web/)), Linkwarden also saves a copy of each webpage as a Screenshot, PDF, and Epub, ensuring accessibility even if the original content is no longer available.
 
 Additionally, Linkwarden is designed with collaboration in mind, sharing links with the public and/or allowing multiple users to work together seamlessly.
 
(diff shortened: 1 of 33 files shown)


daniel31x13:
Hi @rutkai, thanks for your contribution! Archiving as Epub wasn't initially planned, but it's an interesting direction. We also have a request for SingleFile archiving. Could your work be adapted to fit this instead? Your input would be valuable.

daniel31x13:
Hi @rutkai, thanks for your contribution! We hadn't planned on Epub archiving mainly because there hasn't been much demand for it. But we do have a request for [SingleFile](https://github.com/gildas-lormeau/SingleFile) archiving. Could your work be adapted to fit this? Your input would be valuable.

rutkai:
Hi @daniel31x13, sure, implementing single file shouldn't be an issue.

Speaking of Epub, there are tons of tools that forward articles to Kindle (I use such a feature myself too) and I thought this may be a good addition. I'm considering using linkwarden instead of shiori and this feature is missing from here, so I thought why not adding it.

daniel31x13:
I really appreciate the thought you've put into your suggestion - it could be quite useful for specific scenarios like yours. However, we're trying to balance individual ideas with what's best for all our users and the project's path forward. 
Adding too many varied features without a shared vision might make things a bit too complex for most folks.

rutkai:
Understandable. Btw is there a forum to promote my idea? It is possible that others just didn't think of this feature but otherwise would like it to have. Shall I post an issue about it with `enhancement` label?

daniel31x13:
Thank you for your understanding, we already stated how we go for new suggestions over [here](https://github.com/linkwarden/linkwarden#suggestions). Basically yes, an issue with an enhancement label.

Jiar:
> Hi @daniel31x13, sure, implementing single file shouldn't be an issue.
> 
> Speaking of Epub, there are tons of tools that forward articles to Kindle (I use such a feature myself too) and I thought this may be a good addition. I'm considering using linkwarden instead of shiori and this feature is missing from here, so I thought why not adding it.

Same as you 👍

daniel31x13:
There are many conflicting changes in this PR, and also it's not in the scope of this project (as of now) since Linkwarden primarily focuses on archiving as the base formats available. So that the user can convert that format to any other formats...

PR #513, opened 2024-03-11
Update docker-compose.yml - link /data to map /data/archives and /data/uploads

davidosomething:
link the entire data dir

otherwise /data/archives and /data/uploads goes into docker.img ! my linkwarden docker container was taking over 2 GB because the archives were not mapped

image

`du` in the container
image


davidosomething:
never mind it was 1gb of `node_modules`

PR #598, opened 2024-05-08
Fix bookmark import issue with missing folder names

LeonKohli:
### Description
This pull request addresses an issue where importing bookmarks from an HTML file fails when a bookmark folder has no name. The current implementation fails the creation of collections for unnamed folders, leading to an infinite loader and no, or missing bookmarks.

### Example
Here's an example of an HTML snippet that demonstrates the issue:

```html
<!DOCTYPE NETSCAPE-Bookmark-file-1>
<!-- This is an automatically generated file.
     It will be read and overwritten.
     DO NOT EDIT! -->
<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
<TITLE>Bookmarks</TITLE>
<H1>Bookmarks</H1>
<DL><p>
    <DT><H3 ADD_DATE="1715187970" LAST_MODIFIED="0" PERSONAL_TOOLBAR_FOLDER="true">Bookmarks bar</H3>
    <DL><p>
        <DT><H3 ADD_DATE="1715188017" LAST_MODIFIED="0"></H3>
        <DL><p>
            <DT><H3 ADD_DATE="1715188017" LAST_MODIFIED="0">Photos</H3>
            <DL><p>
                <DT><A HREF="https://www.istockphoto.com/de" ADD_DATE="1690043474">Bilder Kaufen, Stockfotos, Vektorbilder, Footage - iStock</A>
                ...
```

In this example, the bookmark folder with the name "Photos" is nested inside an unnamed folder. The current implementation fails to import all bookmarks and creates and infinite loader and wont import any bookmarks up to this point in the html.

### Solution
The proposed solution is to handle the case when a bookmark folder has no name by assigning a default name, such as "Untitled Collection". This ensures that the collection is still created and all bookmarks will be imported correctly.

The changes are made in the `processBookmarks` function in the `importFromHTMLFile.ts` file. Here's the relevant code snippet:

```typescript
if (collectionName) {
  const collectionNameContent = (collectionName.children[0] as TextNode)?.content;
  if (collectionNameContent) {
    collectionId = await createCollection(
      userId,
      collectionNameContent,
      parentCollectionId
    );
  } else {
    // Handle the case when the collection name is empty
    collectionId = await createCollection(
      userId,
      "Untitled Collection",
      parentCollectionId
    );
  }
}
```

When the `collectionNameContent` is falsy (i.e., the folder has no name), a collection with the name "Untitled Collection" is created instead of failing the collection creation altogether.

### Benefits
- Improves the reliability of importing bookmarks from HTML files.
- Ensures that all bookmarks are imported, even if they are inside unnamed folders.

### DIFF
--- a/lib/api/controllers/migration/importFromHTMLFile.ts
+++ b/lib/api/controllers/migration/importFromHTMLFile.ts
@@ -63,11 +63,21 @@ async function processBookmarks(
         ) as Element;
 
         if (collectionName) {
-          collectionId = await createCollection(
-            userId,
-            (collectionName.children[0] as TextNode).content,
-            parentCollectionId
-          );
+          const collectionNameContent = (collectionName.children[0] as TextNode)?.content;
+          if (collectionNameContent) {
+            collectionId = await createCollection(
+              userId,
+              collectionNameContent,
+              parentCollectionId
+            );
+          } else {
+            // Handle the case when the collection name is empty
+            collectionId = await createCollection(
+              userId,
+              "Untitled Collection",
+              parentCollectionId
+            );
+          }
         }
         await processBookmarks(
           userId,
@@ -264,3 +274,4 @@ function processNodes(nodes: Node[]) {
   nodes.forEach(findAndProcessDL);
   return nodes;
 }
+


daniel31x13:
Hello sorry for the late reply and thank you for the PR, gonna merge this soon after I implement the upcoming features.

PR #613, opened 2024-06-03
Add support for local Cookies file

Eisaichen:
When creating preserved formats, a lot of sites will ask for login and create a big pop-up window for it, a big chunk of the website could be covered by that window like the following example:
image

To avoid that, the new code will load a cookies file at /data/data/cookies.txt, which is the data folder for the container image.
So, when taking a snapshot, the playwright will log in with cookies automatically to avoid pop-up for login. when set up properly could also avoid pop-up for cookies permissions and all sorts of other things as well.
cookies file needs to be in the standard format that can be extracted by this extension for Chrome/Edge:
[Get cookies.txt LOCALLY](https://chromewebstore.google.com/detail/cclelndahbckbenkjhflpdbgdldlbecc)

Here is the snapshot of the same webpage as the previous example after the cookies are loaded:
image

### DIFF
--- a/lib/api/archiveHandler.ts
+++ b/lib/api/archiveHandler.ts
@@ -10,6 +10,7 @@ import validateUrlSize from "./validateUrlSize";
 import removeFile from "./storage/removeFile";
 import Jimp from "jimp";
 import createFolder from "./storage/createFolder";
+import fs from "fs";
 
 type LinksAndCollectionAndOwner = Link & {
   collection: Collection & {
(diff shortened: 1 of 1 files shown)


daniel31x13:
Heya, so in the next release there is a much more convenient way of dealing with this issue. Basically users can upload files instead, this enables users to capture a screenshot of the desired section locally and then upload it manually to Linkwarden.

For that I'm gonna have to close this PR, let me know if there's anything else.
Also for future contributions, please stay in touch on what you'll be working on (if any) just in case...

Thank you and sorry for the inconvenience cause.

PR #636, opened 2024-07-05
Tags in public collection

bjoerndot:
This PR adds the ability to show tags as "shortcut" for filtering a public collection. In essence it displays a line of all tags that are found within the public collection and let's users click on those tags. Once they click it, the tag will be set as search parameter for the collection and only links with the chosen tag will be displayed. 
This helps navigating external users within a large collection.

It adds another option in the EditCollectionSharingModal, that let's users choose, whether they want to show the tags:
grafik

If the checkbox above is checked and there acutally are tags within the collection, the public page will show all tags of that collection
grafik

The "All" option resets the search and shows all links.


daniel31x13:
This is a great idea!
A couple of issues:
- If you add more tags, it goes past the screen (first pic).
- If you add the same tag to multiple links, it shows duplicate tags (first pic).
- If you click on each tags, the tags for the other links get removed (second pic).

Anyway I'd appreciate if you can get to these otherwise I'm gonna have to manage it for later.
Also thanks for the contribution!

Screenshot 2024-07-07 at 7 20 10 PM
Screenshot 2024-07-07 at 7 27 15 PM

bjoerndot:
I found solutions to the problems you brought up, @daniel31x13.

The tags won't overflow the screen anymore. Now they break into the next line: 

grafik

Since this behaviour can lead to a very long list (especially on mobile) I added a disclosure around the tags. This disclosure is open by default for each collection. Once the user interacted with the disclosure the state for that disclosure in that collection is stored in localStorage (just as it is done in `Sidebar.tsx`). 
grafik
grafik

In order to retrieve the links that belong to a collection and that don't get removed once one of the tags is clicked (as you described above) I added the `allLinksOfCollection` field in the LinksStore. Everytime the hook for retreiving links is used, this field get populated as well (see `useLinks.tsx`). I am not a Zustand-expert, so I am not sure whether my implementation is best practice.

daniel31x13:
Hey sorry for the late response, there are a lot of features along the way and I came to this afterwards.
Could you please resolve the merge conflicts? Sorry for the inconvenience caused...

bjoerndot:
It took me a while to find a slot... but now it's done. Since much of the codebase changed, since I first created the PR I had to rework how the tags in public collections are retrieved.

daniel31x13:
@bjoerndot huge thanks for your time, I did some modifications like removing the disclosure and the `tagsArePublic` field since I thought they were unnecessary (and would just worsen the user experience) and merged everything to the dev branch.
Let me know what you think when you had the time! :)

PR #637, opened 2024-07-07
Docker Image Rework

### LINKED ISSUE #629: Docker image is seriously oversized
**Describe the bug**
The 2.5.3 docker image size is 2.6 GB. There is a serious oversize problem.

I run several web services on my nas, through Docker containers, and none is so heavy.

theAkito:
Close #629.

Completely rework Dockerfile.
Focusing on size efficiency.

@daniel31x13

Would you like to test?

I noticed, that I cannot login into Linkwarden at all, since the login form never appears, with the newest image. It also happens with the original image, so it's not related to this rework.

I'm also not sure, whether Playwright works. Not sure, what to test, to make sure, it works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,57 @@
-FROM node:18.18-bullseye-slim
-
-ARG DEBIAN_FRONTEND=noninteractive
-
-RUN mkdir /data
+# Made by Akito <the@akito.ooo>
+FROM node:18.18.2-alpine3.18 AS build
 
 WORKDIR /data
 
 COPY ./package.json ./yarn.lock ./playwright.config.ts ./
 
-RUN --mount=type=cache,sharing=locked,target=/usr/local/share/.cache/yarn yarn install --network-timeout 10000000
+RUN \
+  --mount=type=cache,sharing=locked,target=/usr/local/share/.cache/yarn \
+  yarn install --network-timeout 10000000
 
-RUN apt-get update
-
-RUN apt-get install -y \
-    build-essential \
+RUN \
+  apk update && \
+  apk add --no-cache \
+    musl-dev \
+    pkgconfig \
     curl \
-    libssl-dev \
-    pkg-config
-
-RUN apt-get update
-
-RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
-
-ENV PATH="/root/.cargo/bin:${PATH}"
-
-RUN cargo install monolith
-
-RUN npx playwright install-deps && \
-    apt-get clean && \
-    yarn cache clean
-
-RUN yarn playwright install
+    gcc \
+    chromium \
+    nss \
+    freetype \
+    ttf-freefont \
+    libressl-dev && \
+  curl –proto ‘=https’ –tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y && \
+  source $HOME/.cargo/env && \
+  rustup target add x86_64-unknown-linux-musl && \
+  export PATH="/root/.cargo/bin:${PATH}" && \
+  cargo install monolith && \
+  yarn playwright install
 
 COPY . .
 
-RUN yarn prisma generate && \
-    yarn build
+RUN \
+  yarn prisma generate && \
+  yarn build
 
-CMD yarn prisma migrate deploy && yarn start
+FROM node:18.18.2-alpine3.18
+
+ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
+
+WORKDIR /data
+
+COPY --from=build /data .
+
+RUN \
+  apk update && \
+  apk add --no-cache \
+    chromium \
+    nss \
+    freetype \
+    ttf-freefont && \
+  chown -R node:node /data
+
+USER node
+
+ENTRYPOINT [ "/bin/sh", "-c" ]
+CMD [ "yarn prisma migrate deploy && yarn start" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,8 @@ FROM node:18.18.2-alpine3.18 AS build
 
 WORKDIR /data
 
-COPY ./package.json ./yarn.lock ./playwright.config.ts ./
+COPY --chown=node ./package.json ./yarn.lock ./playwright.config.ts ./
+COPY --chown=node docker/local.conf /etc/fonts/local.conf
 
 RUN \
   --mount=type=cache,sharing=locked,target=/usr/local/share/.cache/yarn \
@@ -17,6 +18,8 @@ RUN \
     curl \
     gcc \
     chromium \
+    chromium-swiftshader \
+    font-noto-emoji \
     nss \
     freetype \
     ttf-freefont \
@@ -36,19 +39,22 @@ RUN \
 
 FROM node:18.18.2-alpine3.18
 
-ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
-
 WORKDIR /data
 
-COPY --from=build /data .
+COPY --chown=node --from=build /data .
+COPY --chown=node docker/local.conf /etc/fonts/local.conf
 
 RUN \
   apk update && \
   apk add --no-cache \
     chromium \
+    chromium-swiftshader \
+    font-noto-emoji \
     nss \
     freetype \
     ttf-freefont && \
+  rm -fr /var/cache/* && \
+  fc-cache -f && \
   chown -R node:node /data
 
 USER node
--- a/docker/local.conf
+++ b/docker/local.conf
@@ -1,0 +1,28 @@
+<?xml version="1.0"?>
+<!DOCTYPE fontconfig SYSTEM "fonts.dtd">
+<fontconfig>
+
+ <alias>
+   <family>sans-serif</family>
+   <prefer>
+     <family>Noto Color Emoji</family>
+     <family>Noto Emoji</family>
+   </prefer>
+ </alias>
+
+ <alias>
+   <family>serif</family>
+   <prefer>
+     <family>Noto Color Emoji</family>
+     <family>Noto Emoji</family>
+   </prefer>
+ </alias>
+
+ <alias>
+  <family>monospace</family>
+  <prefer>
+    <family>Noto Color Emoji</family>
+    <family>Noto Emoji</family>
+   </prefer>
+ </alias>
+</fontconfig>


theAkito:
I just noticed, the size problem is still significant.

```
akito13/linkwarden:original 4.36GB # Dockerfile, before these changes.
akito13/linkwarden:test1 2.26GB    # Dockerfile, after these changes.
akito13/linkwarden:test 4.33GB     # Dockerfile with Alpine, but no multi-stage build.
```

The problem is, that further size reduction must be done on JavaScript and `yarn` level.

wlofman021 (approved):

winston0410:
@theAkito don't have answer to your problem yet, but maybe try inspecting the size of each layer with `dive`?

daniel31x13:
@theAkito Just built the container and seems like Playwright isn't working, which means webpages aren't being archived...
```

**Comment by theAkito** [comment]
> @theAkito don't have answer to your problem yet, but maybe try inspecting the size of each layer with `dive`? https://github.com/wagoodman/dive

Yes, already in the progress of doing that, however I'm currently busy and will probably continue end of this week.



> @theAkito Just built the container and seems like Playwright isn't working, which means webpages aren't being archived...
> 
> ```
> linkwarden-linkwarden-1  | [1] ╔═════════════════════════════════════════════════════════════════════════╗
> linkwarden-linkwarden-1  | [1] ║ Looks like Playwright Test or Playwright was just installed or updated. ║
> linkwarden-linkwarden-1  | [1] ║ Please run the following command to download new browsers:              ║
> linkwarden-linkwarden-1  | [1] ║                                                                         ║
> linkwarden-linkwarden-1  | [1] ║     yarn playwright install                                             ║
> linkwarden-linkwarden-1  | [1] ║                                                                         ║
> linkwarden-linkwarden-1  | [1] ║ <3 Playwright Team                                                      ║
> linkwarden-linkwarden-1  | [1] ╚═════════════════════════════════════════════════════════════════════════╝
> ```

Okay, in the worst case scenario, I can switch back to Debian, as `playwright` sadly has no official support for Alpine, however this will make the image even larger.
I had researched, that there may be possibilities to make it run on Alpine, though.

____

The biggest problem right now is the size due to all the `node_modules` stuff, etc. It'd be great, if you could look into that and check, whether there's a way to fix dependencies to some degree, so they are not taking as much space. Not sure, what tools there are beyond replacing/removing dependencies.

From what I gathered online about this topic, it's a general problem, that these builds are way too big, however, compared to all examples online, Linkwarden is still way bigger than whatever I had found online. Therefore, it seems like, there is lots of room for improvement regarding dependency management and size optimisation.

**Comment by winston0410** [comment]
> > @theAkito don't have answer to your problem yet, but maybe try inspecting the size of each layer with `dive`? https://github.com/wagoodman/dive
> 
> Yes, already in the progress of doing that, however I'm currently busy and will probably continue end of this week.
> 
> > @theAkito Just built the container and seems like Playwright isn't working, which means webpages aren't being archived...
> > ```
> > linkwarden-linkwarden-1  | [1] ╔═════════════════════════════════════════════════════════════════════════╗
> > linkwarden-linkwarden-1  | [1] ║ Looks like Playwright Test or Playwright was just installed or updated. ║
> > linkwarden-linkwarden-1  | [1] ║ Please run the following command to download new browsers:              ║
> > linkwarden-linkwarden-1  | [1] ║                                                                         ║
> > linkwarden-linkwarden-1  | [1] ║     yarn playwright install                                             ║
> > linkwarden-linkwarden-1  | [1] ║                                                                         ║
> > linkwarden-linkwarden-1  | [1] ║ <3 Playwright Team                                                      ║
> > linkwarden-linkwarden-1  | [1] ╚═════════════════════════════════════════════════════════════════════════╝
> > ```
> 
> Okay, in the worst case scenario, I can switch back to Debian, as `playwright` sadly has no official support for Alpine, however this will make the image even larger. I had researched, that there may be possibilities to make it run on Alpine, though.
> 
> The biggest problem right now is the size due to all the `node_modules` stuff, etc. It'd be great, if you could look into that and check, whether there's a way to fix dependencies to some degree, so they are not taking as much space. Not sure, what tools there are beyond replacing/removing dependencies.
> 
> From what I gathered online about this topic, it's a general problem, that these builds are way too big, however, compared to all examples online, Linkwarden is still way bigger than whatever I had found online. Therefore, it seems like, there is lots of room for improvement regarding dependency management and size optimisation.

few things come into mind:
- replacing react with preact
- alias `lodash` with `lodash-es`, its ESM equivalent to enable treeshaking. I notice linkwarden does not use it, but a lot of package would use `lodash`, so it should improve performance to a certain extent.

And something more radical but effective might be:
- migrate away from Nextjs. I think a lot of the bloat come from the fact that we need to execute the project with `next`, `concurrently`, `ts-node`, `yarn` and execute Typescript, instead of compiled Javascript.
- migrate away from React to Svelte(Nextjs to SvelteKit). I have batted with bundle size and lighthouse score, Svelte always win over React and with a much smaller footprint.

**Comment by theAkito** [comment]
> > > @theAkito don't have answer to your problem yet, but maybe try inspecting the size of each layer with `dive`? https://github.com/wagoodman/dive
> > 
> > 
> > Yes, already in the progress of doing that, however I'm currently busy and will probably continue end of this week.
> > > @theAkito Just built the container and seems like Playwright isn't working, which means webpages aren't being archived...
> > > ```
> > > linkwarden-linkwarden-1  | [1] ╔═════════════════════════════════════════════════════════════════════════╗
> > > linkwarden-linkwarden-1  | [1] ║ Looks like Playwright Test or Playwright was just installed or updated. ║
> > > linkwarden-linkwarden-1  | [1] ║ Please run the following command to download new browsers:              ║
> > > linkwarden-linkwarden-1  | [1] ║                                                                         ║
> > > linkwarden-linkwarden-1  | [1] ║     yarn playwright install                                             ║
> > > linkwarden-linkwarden-1  | [1] ║                                                                         ║
> > > linkwarden-linkwarden-1  | [1] ║ <3 Playwright Team                                                      ║
> > > linkwarden-linkwarden-1  | [1] ╚═════════════════════════════════════════════════════════════════════════╝
> > > ```
> > 
> > 
> > Okay, in the worst case scenario, I can switch back to Debian, as `playwright` sadly has no official support for Alpine, however this will make the image even larger. I had researched, that there may be possibilities to make it run on Alpine, though.
> > The biggest problem right now is the size due to all the `node_modules` stuff, etc. It'd be great, if you could look into that and check, whether there's a way to fix dependencies to some degree, so they are not taking as much space. Not sure, what tools there are beyond replacing/removing dependencies.
> > From what I gathered online about this topic, it's a general problem, that these builds are way too big, however, compared to all examples online, Linkwarden is still way bigger than whatever I had found online. Therefore, it seems like, there is lots of room for improvement regarding dependency management and size optimisation.
> 
> few things come into mind:
> 
>     * replacing react with preact
> 
>     * alias `lodash` with `lodash-es`, its ESM equivalent to enable treeshaking. I notice linkwarden does not use it, but a lot of package would use `lodash`, so it should improve performance to a certain extent.
> 
> 
> And something more radical but effective might be:
> 
>     * migrate away from Nextjs. I think a lot of the bloat come from the fact that we need to execute the project(yarn start) with `next`, `concurrently`, `ts-node`, `yarn` and execute Typescript, instead of compiled Javascript. All of these add up the size
> 
>     * migrate away from React to Svelte(Nextjs to SvelteKit). I have batted with bundle size and lighthouse score, Svelte always win over React and with a much smaller footprint.

It'd be also nice, if there was a differentiation between production and development dependencies. Some few local tests didn't work out yet, with the `--production` flag.

**Comment by daniel31x13** [comment]
@winston0410 is there a better solution that you could come up with?

> replacing react with preact...

> migrate away from Nextjs...

> migrate away from React to Svelte...

**Comment by daniel31x13** [comment]
@theAkito I simplified the dockerfile here, maybe that might help with anyone who's trying to find a way to improve this.

And I'm guessing the main reason of the large build is because of Playwright, it's basically a browser inside the container and it's used for archiving the webpages.

**Comment by theAkito** [comment]
> I simplified the dockerfile here, maybe that might help with anyone who's trying to find a way to improve this.

Unfortunately, this amount of simplification might heavily increase the image size even further, which is already too high, so I think it's important to keep the Docker part of the image as minimal as possible.



> And I'm guessing the main reason of the large build is because of Playwright, it's basically a browser inside the container and it's used for archiving the webpages.

That definitely sounds huge. Is there no smaller & simpler way to achieve the same? 🤔

____

It would be great, if the project could minimise size issues on the JavaScript side in general.

**Comment by daniel31x13** [comment]
> Is there no smaller & simpler way to achieve the same?

Not without compromising the preservation quality. The best way to archive each webpage is to prevent websites from detecting the automated requests, and the best way to minimize the detection is to use an actual browser (but headless) to make these requests.

**Comment by theAkito** [comment]
> > Is there no smaller & simpler way to achieve the same?
> 
> Not without compromising the preservation quality. The best way to archive each webpage is to prevent websites from detecting the automated requests, and the best way to minimize the detection is to use an actual browser (but headless) to make these requests.

Absolutely makes sense, I already guessed it might be something like that.

That said, I need some assistance in optimising the JavaScript stuff. Have been researching ways to minimise the size of such a project for this pull request, but it's quite a huge topic, with lots of branches.

For example, would it make sense to put everything in a `dist/index.js` file? This is a solution, which puts all the necessary code into a single file, making it feel like a binary. I doubt, it would reduce the size much, but I have read somewhere, that it might reduce the size, as it only really contains the code necessary to run and nothing else, while a whole folder might contain additional data.

Another way is to use better `yarn` options. Like, for example, `--production`. Using these options might reduce the size even further.

There are multiple other ways, according to what I read online, however I am not that familiar with this type of project, so not sure what would be a good way.

Although, it seems obvious to me, there is room for improvement regarding the reduction of the project's size, in general.

**Comment by winston0410** [comment]
> @winston0410 is there a better solution that you could come up with?
> 
> > replacing react with preact...
> 
> > migrate away from Nextjs...
> 
> > migrate away from React to Svelte...

What do you mean better? Easier to implement or better in reducing image size?

**Comment by theAkito** [comment]
> What do you mean better? Easier to implement or better in reducing image size?

I'm pretty sure, it meant, that you display the actual implementation of your suggestion. I.e. perhaps create a pull request for those.

**Comment by winston0410** [comment]
> > What do you mean better? Easier to implement or better in reducing image size?
> 
> I'm pretty sure, it meant, that you display the actual implementation of your suggestion. I.e. perhaps create a pull request for those.

Sorry I don't have anytime for that yet, I hope my 2 cent there would be helpful.

**Comment by theAkito** [comment]
I've now pushed the final verdict for the image. This is as far as it can get with the image size from the Docker perspective in terms of optimisation.

To further reduce image size, the project has to be minimised. This is part of the JavaScript development, not containerisation, specifically.

Minimisation of the JavaScript stuff is out of scope of this Pull Request anyway, so it is another story for another request and issue.

My focus here, to reach the conclusion and merge of this request, is to make sure everything, including Playwright, is working on this Alpine based image. Once that is the case, I have accomplished the goal of this Pull Request, which was minimising the Docker image itself.

**Comment by daniel31x13** [comment]
@theAkito it's now working but the size remained untouched:
image

**Comment by theAkito** [comment]
I'll still document the change, so everyone can look it up later.

```
docker image ls | detect columns | where REPOSITORY =~ 'linkwarden' | first 2
┏━━━┳━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━┓
┃ # ┃     REPOSITORY     ┃    TAG    ┃    IMAGE     ┃      ID      ┃ CREATED ┃  SIZE  ┃
┣━━━╋━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━╋━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━╋━━━━━━━━━╋━━━━━━━━┫
┃ 0 ┃ akito13/linkwarden ┃ test-phat ┃ f30eb0c2a9c6 ┃ f30eb0c2a9c6 ┃ About   ┃ 4.36GB ┃
┃ 1 ┃ akito13/linkwarden ┃ test6     ┃ 8ef03f56225b ┃ 8ef03f56225b ┃ 15      ┃ 2.28GB ┃
┗━━━┻━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━┻━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━┻━━━━━━━━━┻━━━━━━━━┛
```

The first one, is the original one. The last one is the one with all changes applied, which are also contained in this pull request.

**Comment by theAkito** [comment]
@daniel31x13

Are we ready to merge?

**Comment by daniel31x13** [comment]
@theAkito I just built the image, got the same thing as here: https://github.com/linkwarden/linkwarden/pull/637#issuecomment-2214577495

**Comment by theAkito** [comment]
> @theAkito I just built the image, got the same thing as here: https://github.com/linkwarden/linkwarden/pull/637#issuecomment-2214577495

I'll look into it!

**Comment by theAkito** [comment]
@daniel31x13

I have two issues with testing.

1. I cannot log in. `/login` only allows sign up. Sign up works, but I can never log in. Therefore, I cannot test this.
2. I never see the error message from https://github.com/linkwarden/linkwarden/pull/637#issuecomment-2214577495. Do I need to be logged in to be able to see it, if it does not work? How do you view it and therefore check, whether Playwright works?

I need to be able to test all that, to properly fix the image.

**Comment by Ovsyanka** [comment]
> 1. I cannot log in. `/login` only allows sign up. Sign up works, but I can never log in. Therefore, I cannot test this.

Did you try to set `NEXT_PUBLIC_CREDENTIALS_ENABLED=true` in the `.env` file as suggested in #560? Does you issue looks the same as on the screenshot there?

**Comment by theAkito** [comment]
> > 1. I cannot log in. `/login` only allows sign up. Sign up works, but I can never log in. Therefore, I cannot test this.
> 
> Did you try to set `NEXT_PUBLIC_CREDENTIALS_ENABLED=true` in the `.env` file as suggested in #560? Does you issue looks the same as on the screenshot there?

Looks the same, but I don't know when I will have dedicated time to continue testing this. If anyone needs help on getting this pull request running locally, I can answer questions.

File: Dockerfile [hunk 1]
```diff
@@ -1,40 +1,63 @@
...
+COPY --chown=node docker/local.conf /etc/fonts/local.conf
 
-RUN apt-get update
+RUN \
+  --mount=type=cache,sharing=locked,target=/usr/local/share/.cache/yarn \
+  yarn install --network-timeout 10000000
 
-RUN apt-get install -y \
-    build-essential \
+RUN \
```
Ovsyanka, line 13: I believe it would be beneficial to build time to use cache mount here and on the next apk packages installation layer and remove --no-cache. Because some packages installed in both layers and the cache could be used for that.
I didn't test it, but should work I believe:

```
--mount=id=alpine3.18_apk_lib,target=/var/lib/apk,type=cache,sharing=locked \
--mount=id=alpine3.18_apk_cache,target=/var/cache/apk,type=cache,sharing=locked \
```

theAkito: Okay, should be tested. I did not need this type of caching before, so am not that familiar with it.

Ovsyanka:
I run it locally and the only thing I noticed, that playwright still doesn't work. It shows the error in the logs:
`Error processing link https://REDACTED for user 1: browserType.launch: Executable doesn't exist at /home/node/.cache/ms-playwright/chromium-1124/chrome-linux/chrome`

I believe there it isn't worth it to try to use alpine for this purpose because the savings in space between alpine and debian based images isn't very noticeable in comparison to overall image size.

Just for the record, your last image (builded locally, commit b8267cd447ac24466a9338605eb22cb961cd0c6d) is 1.51GB.

daniel31x13:
Hey thanks for your work but we went after #688 since it was less error prone.
Sorry for the inconvenience.

PR #671, opened 2024-07-25
handle undefined

jlssmt:
undefined should be handled as this is accessible by api.


daniel31x13:
Thanks for pointing this out 👍

Issue #661, opened 2024-07-24
NextJS Client Side Exception (maximum call stack size exceeded)


GoodM4ven:
**Describe the bug**
If I try to trigger the dropdown that lists **all** the collections to move the current link(s) to, the whole app crashes really! I'm using lots of indented collections and lots of collections too, so I guess the **complex** mapping of those breadcrumbs (for each option in the select input), plus the many options there are in the first place... -Where I think they shouldn't be even displayed fully to be honest, but rather just a limited amount, plus the ones that would be queried (and also top findings and limited too). Anyway...

![image](https://github.com/user-attachments/assets/40919316-edea-49b5-be27-699278be8b3c)
![image](https://github.com/user-attachments/assets/b6f57478-1854-4a8b-832d-110f32654c72)
![image](https://github.com/user-attachments/assets/61e42479-d92c-4320-ae40-ce7caec30a98)
![image](https://github.com/user-attachments/assets/b39793c8-8bc9-43c9-9cfd-049e4817c36d)

**Software Versions:**
 - OS: Ubuntu 24.04 LTS
 - Browser: Brave 126.1.67.123
 - App: LinkWarden 2.6.0

*Feel free to request any additional details from me*

GoodM4ven:
It still throws it in v2.7.1

![image](https://github.com/user-attachments/assets/5294f6c4-0031-4355-ba7b-2d343d1bb539)

IsaacWise06:
It crashes when you click the dropdown to move a link to another collection?

GoodM4ven:
> It crashes when you click the dropdown to move a link to another collection?

Yes. It does. Where are the logs exactly, so I could provide you if it helps?

PR #678, opened 2024-07-27
General Fixes

IsaacWise06:
- Replace ternary operators with logical ANDs 
- Replace `any` types with the correct type
- Redact `parentId`, `collectionId`, and `ownerId` when exporting data


GoodM4ven:
Greetings,

Can you confirm whether this is necessary for collection-dropdown processing or not:

[CollectionSelection.tsx]
```
const getParentNames = (parentId: number): string[] => {
  const parentNames = [];
  const parent = collections.find((e) => e.id === parentId);

  if (!parent) return parentNames; // return early if parent not found

  parentNames.push(parent.name);
  if (parent.parentId) {
    parentNames.push(...getParentNames(parent.parentId));
  }

  return parentNames.reverse();
};
```

Cause it seems to me that it causes infinite recursion for the top level collection! -Or something related to [this](https://github.com/linkwarden/linkwarden/issues/661) issue.

daniel31x13:
@IsaacWise06 Hey can you fix the merge conflicts whenever you had a chance? The conflicts are mainly due to my changes to the whole state management... My bad :P

PR #689, opened 2024-08-05
Chore: Updates actions to their latest versions

stumpylog:
No problems building or testing:
- https://github.com/stumpylog/linkwarden/actions/runs/10239661604
- https://github.com/stumpylog/linkwarden/actions/runs/10241140570

Will eliminate the current warnings.

If helpful, I could configure dependabot for monthly checks of action updates


daniel31x13:
Thanks for the PR! I think this is good for now, we can just manually update as soon as it gives us warnings...

PR #692, opened 2024-08-07
Update [...nextauth].ts to allow existing SSO user sign

### LINKED ISSUE #687: DISABLE_NEW_SSO_USERS = true blocks SSO use for existing users, but allows new users
**Describe the bug**
DISABLE_NEW_SSO_USERS = true blocks SSO use for existing users, but allows new users

**To Reproduce**
Set DISABLE_NEW_SSO_USERS = true
Log in with SSO existing user and get Access Denied message
Log in with SSO new user, and a new user is created and logged in

**Expected behavior**
New users should be blocked but existing users should be allowed

**Screenshots**
Screenshot 2024-08-03 112258

Docker Container

phillibl:
Fixed issue where sign in would fail for existing user if DISABLE_NEW_SSO_USERS = true

Resolves #687

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,18 @@
 <div align="center">
   <img src="./assets/logo.png" width="100px" />
   <h1>Linkwarden</h1>
+  <h3>Bookmark Preservation for Individuals and Teams</h3>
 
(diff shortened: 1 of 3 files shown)


daniel31x13:
Thanks for the contribution!

PR #710, opened 2024-08-17
Please 'EXPOSE' port 3000

arran4:
This is so I can setup up Synology correctly


daniel31x13:
Can we set this dynamically? Like what if a user wanted to host this on port 3001?

arran4:
This is really just meta data to help the user (or in this case systems like Container Manager on synology.) The user manually specifies the port with the `--publish` option regardless, this just lets the user know what is exported from the container. If you have `EXPOSE 3000` here and run `docker run --publish-all` this will cause docker to effectively run `docker run --publish 3000:3000` 

Documentation is here: https://docs.docker.com/engine/containers/run/#exposed-ports

For other Synology users: At the moment I managed to get it working by exporting the image, manually modifying the application data and uploading it back: You need to set:
```
   "enable_service_portal" : true,
   "service_portals" : [
      {
         "port" : "3000",
         "protocol" : "http"
      }
   ],
   "services" : [
      {
         "display_name" : "linkwarden-1 (port: 3000)",
         "id" : "Docker-$DOCKERIDHERE-3000",
         "proxy_target" : "http://172.17.0.10:3000",
         "service" : "Docker-$DOCKERIDHERE-3000",
         "type" : "reverse_proxy"
      }
   ],
```

Replace `$DOCKERIDHERE` with the contents of `id`

daniel31x13:
Sorry for the late response, merging this for the next release.

arran4:
Thanks

PR #726, opened 2024-08-19
Add Configurable Playwright Concurrency via Environment Variable

CoffeeAnon:


CoffeeAnon:
Minor change to add an optional env var to set a max number of workers.

This is relevant in k8s environments where resource limits can be set but large numbers of parallel tasks will trigger an OOM kill.

PR #800, opened 2024-10-21
Remove duplicated URL if existing (MacOS Shortcut), fixes #551

### LINKED ISSUE #551: Safari Extension / Working iOS shortcut on macOS
I have attempted to adapt the iOS Shortcut to use on macOS (since there isn't an extension for Safari, my primary browser on my Mac. 

However, when I submit a URL via the same Shortcut on macOS (that works great on iOS) it enters the URL twice into the URL field of the saved link.

So for example using the same Shortcut: 
- On iOS it results in a saved link with the URL `www.github.com`
- Whereas on macOS it results in a link `www.github.comwww.github.com` (which obviously ends up in a 404 when clicked)

Not sure what is going on but would love to use this with Safari on my Mac! 

Thanks for your awesomeness!

andreasfaerber:
Remove everything after a \n in the url via api submission. Please check the fix as my typescript skills are non-existent.

### DIFF
--- a/lib/api/controllers/links/postLink.ts
+++ b/lib/api/controllers/links/postLink.ts
@@ -10,6 +10,10 @@ export default async function postLink(
   link: LinkIncludingShortenedCollectionAndTags,
   userId: number
 ) {
+
+  const trimUrl = link.url?.replace(/\n.*$/, "") || null;
+  link.url = trimUrl;
+
   if (link.url || link.type === "url") {
     try {
       new URL(link.url || "");


daniel31x13:
Thanks for the contribution! This seems a hacky workaround, the problem is the shortcut itself. It shouldn't be duplicating the url, Idk why that is happening on Mac.

hjtbuitenhuis:
It is a hacky workaround, but it appears this is a bug in OSX. 

I have personally resolved it by modifying the shortcut to take the first item from the list:
image

pablosed:
> It is a hacky workaround, but it appears this is a bug in OSX.
> 
> I have personally resolved it by modifying the shortcut to take the first item from the list: image

I was also having this issue, and this solved it. Thank you for making the suggestion!

PR #809, opened 2024-11-01
update german translation

Green-Kite:
German translation updated as mentioned by @daniel31x13 in #711.

Not sure if I fetched all changes.

PR #815, opened 2024-11-03
update german translation

Green-Kite:
@daniel31x13 updated german translation with missing fields from #809.


daniel31x13:
Thanks!

PR #847, opened 2024-11-13
Switch PostgreSQL container to use pgautoupgrade

Tealk:
This handles automatically upgrading the data files to newer versions of PostgreSQL.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.5"
 services:
   postgres:
-    image: postgres:16-alpine
+    image: pgautoupgrade/pgautoupgrade:16-alpine
     env_file: .env
     restart: always
     volumes:


daniel31x13:
This isn't a necessary thing to implement for now as it's an extra layer of abstraction when the default postgres image is already doing the job well.
Also what if there was a breaking change in one of the major versions?

Tealk:
I don't understand what you mean, if you change the docker image from 16 to 17, then you have to manually extract the DB of 16, update it to 17 and import the DB again.

This package does these steps automatically.

daniel31x13:
Maybe it's just me but I think pgautoupgrade isn't a well know package (never read of it). And it might cause some bugs on the long run.

If you think it's really a nice thing to have the most cutting-edge version of Postgres, feel free to make the pull request and we'll see if it get's any traction... (gonna merge it if the community wants)

Tealk:
It's more a matter of at some point there is an eol and then you have to update. And then you have the manual effort described above. The image I suggested would make it easier and you wouldn't need to have expertise in PosqreSQL to perform an update.

Issue #534, opened 2024-03-21
Reverse proxy into a basePath


Yarg-mirror:
Hello,

For reverse proxying, instead of the usual https://<linkwarden>.example.com in a sub domain, I am more used to run most of my services with an url of the type: https://services.example.com/<linkwarden>

I was wondering if linkwarden was compatible with such configuration and how to configure it for to take that into account? My worry is that some application use to hardcode some link (like for example /api) that will reset the basePath and call for https://services.example.com/api/... instead of https://services.example.com/<linkwarden>/api/...

Some time there is some workaround to make it more or less work, but other time it's impossible to do if it not something that can be configured.

**additionnal informations:**
For the reverse proxy part I plan to use HAProxy.

Thanks in advance for your help! I hope my explanation isn't too messy.

zudljk:
I'm not entirely sure what you mean ... https://services.example.com/api and https://services.example.com//api are more or less equivalent, reverse proxies should handle that.

Do you want to do something like hosting linkwarden under a base URL that includes a path, like https://services.example.com/linkwarden/...?

Maybe the maintainers can clear up on this but to my knowledge, that's not possible. For this to work, a web application has to support setting the basePath somewhere, since it needs render it into all the internal links it generates. There's currently no such setting, at least I haven't found any in the docs or source code.

Yarg-mirror:
That's exactly it, sorry about that, I messed my post because I used a few < in the examples urls, that resulted in them now showing

lezli01:
Does linkwarden support changing its basepath? So it could be put behind a reverse proxy under some custom resource?

kasisoft:
I've stumbled upon this issue myself and created a pull request https://github.com/linkwarden/linkwarden/pull/872 (which requires to be added by the maintainers unless they reject it).

Yarg-mirror:
Thank you for your work, that is a great news! I hope it will be accepted :)

PR #874, opened 2024-11-19
534: Provide a basePath in support of Proxies

kasisoft:
Ticket: https://github.com/linkwarden/linkwarden/issues/534

# Changes:
* The _next.config.js_ has been updated to configure the _basePath_ attribute which is generally meant to be used for such cases.
* The _basePath_ which is accessible through the router has been added to (hopefully) all API routes and references except for Next components such as __Link__ which already provide support for it.
* For the following files __getPublicCollectionData.ts__, __getPublicUserData.ts__, __links.ts__ the _basePath_ is taken from the configuration _next.config.js_ as the router isn't available there.

# Configuration:
* If no configuration takes place the behaviour remains as before.
* The _basePath_ can be configured in an _env_ file such as .env, .env.local. Example:
```
BASE_PATH=/linkwarden
```

# Caveats:

* I would have preferred to run and add tests. Unfortunately they didn't run out of the box and fixing these tests might require some time as I'm not a frontend dev.
* The _basePath_ support for the application is straight forward. However the use through Docker/Compose is not as simple. The compilation process requires to have the environment variables accessible during the build. This could be solved if the environment variables could be written into the __.env__ file within the __Dockerfile__ which isn't feasible in a sensible way (at least not for me) so doing this properly would require some additional work. Therefore support for the _basePath_ through Docker/Compose requires to explicitly build the image.

# Documentation: 
I have no idea where to put this as the support for the _basePath_ currently requires explicitly compiling an image (unless someone has a better idea).

Let's create an example folder __/home/dodo/mydir__ (just something apropriate for you) and follow these steps:

```bash
# Get the current repository (-> /home/dodo/mydir/linkwarden)
git clone https://github.com/linkwarden/linkwarden.git
```

Create a file for the configuration __/home/dodo/mydir/linkwarden/.env.local__:
```
# MUST start with a slash and NOT end with a slash
BASE_PATH=/linkwarden
```

Build an image for this. Assume the image name "my-linkwarden-image" you can build it like this:

```bash
docker build -t my-linkwarden-image:1 .
```

If everything went fine you can refer to the image "my-linkwarden-image:1" within the _docker-compose.yml_ .

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,10 @@
 NEXTAUTH_URL=http://localhost:3000/api/v1/auth
 NEXTAUTH_SECRET=
 
+# Deploy Linkwarden using the path prefix '/linkwarden'. Note that
+# this prefix must start with '/' and NOT end with a '/' 
+# BASE_PATH=/linkwarden
+
 # Manual installation database settings
 # Example: DATABASE_URL=postgresql://user:password@localhost:5432/linkwarden
 DATABASE_URL=
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,7 @@ certificates
 /playwright-report/
 /blob-report/
 /playwright/.cache/
+
+# pnpm
+pnpm-lock.yaml
+
--- a/components/LinkDetails.tsx
+++ b/components/LinkDetails.tsx
@@ -219,7 +219,7 @@ export default function LinkDetails({
         >
           {previewAvailable(link) ? (
             <Image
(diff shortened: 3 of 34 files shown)


CatLover01:
Would love to see this merged! It's exactly what I need for my setup.

backelj:
@kasisoft just wondering, why was this closed? any reason why the basePath changes are not accepted?

kasisoft:
Probably because I deleted all my forks on Github as I'm moving away from Github. Don't know why it hasn't been accepted.

Issue #808, opened 2024-10-31
Omnivore import


thiswillbeyourgithub:
Hi,

With omnivore [getting sold to ElevenLabs](https://blog.omnivore.app/p/omnivore-is-joining-elevenlabs) and adding the ability to [fully export our data](https://docs.omnivore.app/using/exporting.html) I was wondering wether it was possible to import their export format into linkwarden?

daniel31x13:
Sure, I just need a sample file. Like what the exported data looks like...

sachinsenal0x64:
> Sure, I just need a sample file. Like what the exported data looks like...

Like this

```json
[ {
    "id": "c2c5dc3e-9bf3-11ee-a9ec-37d11a195906",
    "slug": "asdf-asdf-18c71e973e9",
    "title": "asdf | asdf",
    "description": "Manage multiple runtime versions with a single CLI tool",
    "author": null,
    "url": "https://asdf-vm.com",
    "state": "Active",
    "readingProgress": 0,
    "thumbnail": null,
    "labels": [],
    "savedAt": "2023-12-17T10:03:12.079Z",
    "updatedAt": "2023-12-17T10:03:12.100Z",
    "publishedAt": null
  },
  {
    "id": "c47194a6-9bf3-11ee-a9ec-b71fb3eeb272",
    "slug": "website-optimization-and-digital-agency-sales-tools-woo-rank-18c71e97ee2",
    "title": "Website Optimization and Digital Agency Sales Tools | WooRank",
    "description": "Our easy-to-use SEO tools will help you optimize your online presence, grow your business and generate more leads.",
    "author": null,
    "url": "https://www.woorank.com",
    "state": "Active",
    "readingProgress": 0,
    "thumbnail": "https://assets-global.website-files.com/5f92a84224a1222449ebbe88/6182a3c5a2f9a52d594db16c_OG_WooRank.jpg",
    "labels": [],
    "savedAt": "2023-12-17T10:03:07.930Z",
    "updatedAt": "2023-12-17T10:03:07.968Z",
    "publishedAt": "2023-11-15T00:00:00.000Z"
  },]
```

dereulenspiegel:
Hello :)

I also started some implementation (wanted to play around first to be sure I can do this since I am not a typescript developer by trade) you can see here: https://github.com/dereulenspiegel/linkwarden/tree/omnivore-import

[...]

PR #875, opened 2024-11-19
Omnivore import

dereulenspiegel:
I started implementing imports from Omnivore as I also needed that feature (see #808).

# Changes

* Modified the API `/api/v1/migration` so that the import format is now specified as part of the URL (i.e. `/api/v1/migration/omnivore`). The body is accepted as binary stream which makes handling binary formats like ZIP files (as needed for Omnivore) easier.
* Adopted the existing import functions to accept a `Readable` as parameter for the data.
* Added the `importFromOmnivore` function.
* Added english translation for the string `from_omnivore`.
* Made the function `importBookmarks` and the Import Bookmark Dropdown reusable (as it is used in two places).

# Missing

* [x] Limitiation on how large import uploads can be. Previously this was ensured by the bodyReader, but I had to disable it to enable binary uploads (maybe there is some kind of "limiting stream reader").
* [x] ~~Automated tests~~(I haven't found any unit tests and only a few e2e tests, please point me in the right direction if that is false)

I think especially the API change probably requires a discussion and I can fully understand if such a change is unwanted. But here are my pro arguments:

* Now the API can handle more file formats and should be easy to extend to handle additional file formats.
* We only pass readable streams to the import functions, making optimizations there possible so that not every import has to be read fully into memory.


daniel31x13:
Hey big thanks for the contribution! 

Two things:

1. Instead of uploading the whole ZIP file just to extract the JSON `metadata_` file, why not just ask the user to upload the metadata file itself? We also won't need the packages `@zip.js/zip.js` and `stream-to-blob` that way.
2. We can't modify the API route due to the breaking changes it causes (i.e. `/api/v1/migration/[format]`). Could you please revert it to its previous state?

Sorry for not responding earlier, there was a lot I was doing behind the scenes.

dereulenspiegel:
Hi,

no worries, just glad that you are running a cool open source project here :) 

1. Unfortunately that would make importing quite uncomfortable for the user, because in the ZIP export from Omnivore are multiple metadata files. They seem to batch the export somehow and then ZIP it.
2. Yes, of course I can do that. But I would still like to keep the ability to import binary files. Would a possible solution be to add a new import method to the v2 API?

daniel31x13:
Oh gotcha! I thought there was only one json file, my bad.

> But I would still like to keep the ability to import binary files.

For the zip files? If so, couldn't we do the same thing on the `/api/v1/migration` path?
Honestly, accepting binary files could raise some security concerns.

dereulenspiegel:
Working with ZIP-files on the current path would at least be quite inefficient and cumbersome. Currently import files are just added as text to a json structure. So for a ZIP-file we would probably need to Base64 encode it on the client and decode it on the backend. But that has the implication that now the real achievable import size is always less than what is configured in `IMPORT_LIMIT` because Base64 encoding creates overhead.

Regarding the security concerns I am curious what you have in mind? Yes, we depend very much on the security of the ZIP library, but so do we for every other decoder.

What could be an interesting solution is to decompress the ZIP file in the client, concatenate all the metadata json files and then upload them via the existing API. But now we suddenly have an import API which, strictly speaking, depends on quite a bit of client behavior for preprocessing the actual import file.

What I would prefer to avoid is for the user to manually extract the archive and upload the "correct" parts of the archive (there is more in there than just the metadata files). This sounds like a horrible UX to be honest.

daniel31x13:
Hmm, I'm thinking if we could create a single `/api/v1/uploads` route where we could do all the file uploading so there's only a single route to worry about breaches, because currently we _do_ have an upload route for the file upload functionality:
https://docs.linkwarden.app/api/uploads-a-file-to-an-archive-by-link-id

Otherwise I guess we have to take the client side preprocessing route...

dereulenspiegel:
What about adding a `/api/v1/migration/upload/[format]` route for imports of binary formats (maybe there will be more in the future?).
Then I could revert the existing import mechanisms and use the new route only for omnivore imports right now.

daniel31x13:
Sure that's also a viable option. And btw could you please use `formidable` to do the security checks (forgot about that), we're already doing that in the [upload file](https://github.com/linkwarden/linkwarden/blob/ed7fab04735e77e98c52c150b4973339ed8ab5c8/pages/api/v1/archives/%5BlinkId%5D.ts) api route.

dereulenspiegel:
I implemented the changes mentioned above so far. But these changes still need testing, I hope to get to that after the weekend.
Regarding `formidable` I am not sure what security benefits it brings. It seems to just do multipart parsing, but I don't use multipart uploads on that route since I need just a very simple binary file upload for the omnivore import. The only feature of `formidable` maybe useful in this context is the limitation of the upload file size, but this is taken care of by `stream-size`.

daniel31x13:
> Regarding formidable I am not sure what security benefits it brings. It seems to just do multipart parsing, but I don't use multipart uploads on that route since I need just a very simple binary file upload for the omnivore import.

Well that's the way to do/handle binary uploads.

Regarding the `stream-size` package, we shouldn't be using an [untested piece of code from 4 years ago](https://github.com/Sumolari/stream-size)...
Also next time please first pull all the changes instead of force pushing, it causes merge conflicts.

Thanks and sorry for being too bossy lol.

daniel31x13:
Hey, I'll be handling this PR from now on. Thanks for the work so far!
(Just mentioning this to avoid any duplicate effort.)

dereulenspiegel:
Thank you. I am anyway quite sick and will need a few more days to recover before I could have continued my work here.
Looking forward to seeing this feature in one of the next releases.

PR #893, opened 2024-11-29
Adding healthcheck to docker image

zodac:
Updated the docker image with a HEALTHCHECK directive to avoid the need for users to specify one themselves in a docker-compose.yml (for example).

Used some typical values for the check, happy to tweak if necessary.

Also added the **curl** package (removed in https://github.com/linkwarden/linkwarden/commit/d60200205aa41f10930c932e247fb46e7547ccf6, but needed for the healthcheck to be executed).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,12 @@ COPY . .
 RUN yarn prisma generate && \
     yarn build
 
+HEALTHCHECK --interval=30s \
+            --timeout=5s \
+            --start-period=10s \
+            --retries=3 \
+            CMD [ "/usr/bin/curl", "--silent", "--fail", "http://127.0.0.1:3000/" ]
+
 EXPOSE 3000
 
 CMD yarn prisma migrate deploy && yarn start
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,13 @@ COPY ./package.json ./yarn.lock ./playwright.config.ts ./
 
 RUN --mount=type=cache,sharing=locked,target=/usr/local/share/.cache/yarn \
     set -eux && \
-    yarn install --network-timeout 10000000
+    yarn install --network-timeout 10000000 && \
+    # Install curl for healthcheck
+    apt-get update && \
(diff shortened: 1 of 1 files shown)


mikelococo:
Have you tested this, and if so, how?

I'm just now setting up an explicit health check for my own installation and `﻿﻿﻿﻿﻿﻿ghcr.io/linkwarden/linkwarden:v2.8.4` doesn't appear to have `/usr/bin/curl` available. This confuses me a bit because the upsteam node image has seemingly [had it installed for the last 2y](https://github.com/nodejs/docker-node/blame/59baa0a57dfb356e842b989631077f3c98c934de/18/bullseye-slim/Dockerfile#L20).

For the moment I've fallen back to using bash to check if the port is listening with a healthcheck command like `timeout 10 bash -c 'cat < /dev/null > /dev/tcp/localhost/3000'` but I was hoping to use something that would hit the database like https://docs.linkwarden.app/api/get-user-by-id. I'm a bit confused about the missing curl binary though.

zodac:
Hmm, I see where I went wrong. I didn't build a local linkwarden image to test, but just updated the Dockerfile with the healthcheck that I use in my docker-compose:

```
    healthcheck:
      interval: 30s
      retries: 3
      start_period: 30s
      test: [ "CMD", "/usr/bin/curl", "--silent", "--fail", "http://127.0.0.1:3000/" ]
      timeout: 5s
```

But I'm using v2.7.1. Between that version and the latest, it looks like the curl binary was removed.

It wouldn't be much impact to add it back in (or **wget** which would be smaller if image size is a concern).

mikelococo:
Might be worth adding it back in with this pull request, then.

For my part as a user, I very much support having curl in any image that exposes a webservice. It's extremely useful for both healthchecks and general debugging. I'd also appreciate having this healthcheck built into the image. Hopefully the maintainers will chime in at some point to give feedback or accept the PR.

daniel31x13:
Thanks for the contribution! Merging this now...

PR #902, opened 2024-12-01
fix(import): save pinnedLinks from Linkwarden export json

keizie:
Pin newly imported links if backup.json pinnedLinks has it.

Tested to successfully pinned on my local installation.
Performance might be low if pinnedLinks are numerous. Improvements needed.


daniel31x13:
Thank you, this works for now!

PR #919, opened 2024-12-09
move playwright to dev deps

### LINKED ISSUE #807: Make playwright an optional dependency
**Is your feature request related to a problem? Please describe.**
Microsoft has [officially declared](https://github.com/microsoft/playwright/issues/33327) that they won't support BSD in Playwright, so I can't run Linkwarden on my OpenBSD servers. (No, we don't have Docker. We don't want Docker either.) As there is no technical reason to *require* Playwright just to *use* Linkwarden, I would prefer a solution here.

**Describe the solution you'd like**
`if ($OS contains "BSD") skip_playwright();` (pseudo-code, obviously.)

**Describe alternatives you've considered**
Use something else than Linkwarden.

jonesmac:
fixes #807


daniel31x13:
Hey sorry for the late response, this won't be the fix for #807. I know I told you to make the PR but I misunderstood what you were going to do. My bad.

PR #920, opened 2024-12-09
fix: HTML backup with Monolith when using Docker

Zalaxx:
Since the changes to the new Docker image, HTML backups are not behaving properly: every external resource to the webpage (even on the same domain) are not backed up on the file, which result in a completely broken preview and file.

Basically, the issue is that the previous image was bundling CA certificates, and the new one does not. 
This cause issues to Monolith because it need valid certificates when it is retrieving the resources of a webpage for archving.

I simply add ca-certificates to the image, and keep it lean by cleaning afterwards.


daniel31x13:
Hey thanks for your contribution and sorry for the late reply!
Could you please give me an example link that gets archived better with this image and compare it with the one everyone else is using?
Trying to keep the image size as small as possible...

Zalaxx:
Hi, no worries,
I have the issue with any link really. It fails to fetch any resource (CSS, images...) outside from the base HTML retrieved by playwright and passed down to Monolith.

My docker image is [this one](https://github.com/linkwarden/linkwarden/pkgs/container/linkwarden/314732249?tag=v2.8.4) (same digest, linux/amd64).
I am using podman instead of docker, but I don't think it should and can cause this kind of issue.
I can reproduce the bug multiple time by (re)starting the docker, preserve without `ca-certificates`, notice the issue and then reinstall `ca-certificates` (with `docker exec -it <id> /bin/bash`), refresh the preserved format and have a properly saved HTML version.

As I explained briefly in the issue, I think it makes sense since playwright is using chromium and chromium has [its own ca store ](https://www.chromium.org/Home/chromium-security/root-ca-policy/) bundled with it. It allows every other format to be preserved successfully (since they all depend on the output from playwright). Monolith will fail to retrieve additional resources (but still have a barebone yet valid HTML page) since it has no Root CA by itself, thus failing on HTTPS connection.
Testing with a HTTP only website (like http://http.badssl.com/), the preservation works as expected, with everything retrieved and saved in the HTML file.

Below is an example with https://www.netmeister.org/blog/threat-model-101.html:

Without `ca-certificates`:
image

With `ca-certificates`:
image




I also tried to run monolith from the terminal inside the docker.
```
root@a451637f64c0:/data# monolith https://www.netmeister.org/blog/threat-model-101.html
https://www.netmeister.org/blog/threat-model-101.html (error sending request for url (https://www.netmeister.org/blog/threat-model-101.html))
Could not retrieve target document

root@a451637f64c0:/data# apt install -y ca-certificates
The following additional packages will be installed:
  openssl
The following NEW packages will be installed:
  ca-certificates openssl
0 upgraded, 2 newly installed, 0 to remove and 27 not upgraded.
Need to get 1017 kB of archives.
After this operation, 1898 kB of additional disk space will be used.
... snip ...
done.

root@a451637f64c0:/data# monolith https://www.netmeister.org/blog/threat-model-101.html -o test.html
https://www.netmeister.org/blog/threat-model-101.html
https://www.netmeister.org/blog/threatmodel/slide01.png
... snip ...
https://www.netmeister.org/favicon.ico

root@a451637f64c0:/data#     # The file is properly backed up
```

daniel31x13:
Gotcha thanks! Just one thing, the image size increased from 1.91GB to 3.1GB, is there any way we can reduce the size of the image to save up some space? If not, I'll merge this anyway since it worked out pretty well!

Zalaxx:
Eh, thats weird. I just tried to rebuild locally the image with and without my patch, and there is no difference in size between the two images (2.03GB according to `podman images`). 
When installing `ca-certificates`, apt is reporting a ~2MB increase in space taken, very far from the 1.2GB you're reporting. 
I also checked using [dive](https://github.com/wagoodman/dive), and the new layer is only taking 3.6MB, which is around what I expected from such a small change.

The builds were done by running `podman build . -t <tag> --no-cache`.

daniel31x13:
Strange, then there might be something cached or something on my side...
Merging this now, thanks for your work!

Issue #557, opened 2024-04-08
HTML import breaks URLs with multiple query params


AverageHelper:
**Describe the bug**
URLs that contain multiple query params are broken after importing from `.html`. The `&` characters turn into `&amp;`, completely breaking the URL.

**To Reproduce**
1. Go to any URL with multiple query params, such as https://www.webtoons.com/en/canvas/monsters-and-girls/plague-girlfriends/viewer?title_no=773948&episode_no=1
2. Bookmark the page in your browser
3. Export bookmarks to a `.html` file
4. Import those bookmarks
5. Observe the incorrect URL for the bookmark entry in Linkwarden's UI: https://www.webtoons.com/en/canvas/monsters-and-girls/plague-girlfriends/viewer?title_no=773948&amp;episode_no=1. In Webtoons' case, attempting to go to that URL results in an error page.

**Expected behavior**
Linkwarden should copy the URL exactly how it's stored in the `.html` file, and reproduce that URL as a link in the UI.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Firefox
 - Version 124

Staples1010:
I'm sure this has already been spotted, but just in case it's not:

This bug also seems to affect page anchors - e.g. example.com/slug#section

Staples1010:
> I'm sure this has already been spotted, but just in case it's not:
> 
> This bug also seems to affect page anchors - e.g. example.com/slug#section

Scratch that. It looks like this was caused by a bad export from LinkAce. It seems their export tool didn't handle """ , "'" & "&" very well.

@AverageHelper, Are you able to check if this may be the case with your exported HTML? I've managed to get around this with 'find and replace'.

AverageHelper:
@Staples1010 I'm not sure what exactly I'm looking for, but the only instances of `&amp;` I found were not in any of the `HREF` attributes. (Some webpage titles were formatted badly at the source, not an export issue.)

The URLs in my file appear to be correct. I can copy and paste those URLs from the file directly and the browser navigates to the expected webpage without issue. The trouble seems to be when Linkwarden imports these URLs, they get mangled.

Staples1010:
@AverageHelper, Just tried this again and you're right. I think i wasn't paying attention and used JSON. Followed your instructions with https://www.webtoons.com/en/canvas/monsters-and-girls/plague-girlfriends/viewer?title_no=773948&episode_no=1 and it replaced `&` with `&amp;`.

[...]

PR #942, opened 2024-12-29
[Fix] Work around parser bug that mangles attribute values

AverageHelper:
Addresses #557

It seems `jsdom` rewrites some attributes such that certain characters are replaced with [HTML Entities](https://developer.mozilla.org/en-US/docs/Glossary/Character_reference). For example, a URL bookmarked at `https://example.com/foo&bar=baz` parses as `https://example.com/foo&amp;bar=baz` (note the `&amp;`, which is not the same URL). Other attributes with URLs in them are also read incorrectly, but the user-facing issue is with `HREF` attributes.

This PR uses the same HTML entity encoder (the `entities` package) to decode JSDom's output correctly for both link URLs and link tags.

### DIFF
--- a/lib/api/controllers/migration/importFromHTMLFile.ts
+++ b/lib/api/controllers/migration/importFromHTMLFile.ts
@@ -81,9 +81,9 @@ async function processBookmarks(
       } else if (item.type === "element" && item.tagName === "a") {
         // process link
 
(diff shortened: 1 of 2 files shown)


AverageHelper:
Not sure if changing the base branch broke anything :sweat_smile: but worth a try I guess?

PR #972, opened 2025-01-07
Localization and small fixes

siberianspot:
Fix Russian localization, Correcting a duplicate value in the settings, Adding some missing localizations


daniel31x13:
Hey thanks for the help!

PR #976, opened 2025-01-08
Add format support for rss feed

antonydenyer:
This will allow you to specify `?format=` at the end of the feed.

I'm not confident with this approach for fetching raw content
```
      const response = await fetch(`${protocol}://${host}/api/v1/archives/${link.id}?format=${format}`);
```

when I was testing locally, I hit the 4mb nextjs api limit. Hopefully there's a better approach I'm not aware of.

### DIFF
--- a/pages/public/collections/[id]/rss.tsx
+++ b/pages/public/collections/[id]/rss.tsx
@@ -1,16 +1,21 @@
 import React from "react";
 import { GetServerSideProps } from "next";
 import { prisma } from "@/lib/api/db";
+import { ArchivedFormat } from "@/types/global";
+
 import RSS from "rss";
 
 const RSSFeed: React.FC = () => null;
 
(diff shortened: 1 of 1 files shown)


daniel31x13:
Thanks a lot for the PR and sorry for the late response.

The issue here is that when there is a large amount of readable content (which is often the case), it puts significant stress on the server to generate the feed.

antonydenyer:
> The issue here is that when there is a large amount of readable content (which is often the case), it puts significant stress on the server to generate the feed.

I guess you could make the number of items configurable. Another option is to put a max on the size of each item returned.

daniel31x13:
I mean, we could do a mix of both alongside some sort of caching. That way if users keep refreshing the page, it shows the same content over and over...

That being said, correct me if I'm wrong, but I think including the description in the feed should be sufficient since RSS feeds are meant to provide a glimpse of the content, not the full content.

PR #999, opened 2025-01-14
feat: add grouped links component and related configs

NightWatcher314:
- add Divider and GroupedLinks components
- update SortDropdown and Links components
- add EN/ZH localization configs
- update local settings and global types

Because I used to use onetab, so I want a feature that can show link by group like this.
image

So, I add the group by feature as an switch in dropmenu, It will not affect previous behavior
image

link will group by the sort type, in this picture, it is group by the date.

### DIFF
--- a/components/Divider.tsx
+++ b/components/Divider.tsx
@@ -1,0 +1,7 @@
+export default function Divider() {
+  return (
+    <div className="flex items-center py-2">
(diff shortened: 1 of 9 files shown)


daniel31x13:
If you think otherwise, please open an issue requesting this as an enhancement and over there we can see if people want this feature or not.

PR #1003, opened 2025-01-16
Fix mistranslation and add new translation in Japanese

Kur0den:
I fixed what seemed to be mistranslations, and added keys and translations that were in the English translation file but not in the Japanese translation file.

There were some translations that couldn't be seen in a self-hosted environment, so there may be a few mistranslations  that are out of context.

I also changed some translations to (more exciting) ones.


daniel31x13:
Thank you for your contribution!

PR #1016, opened 2025-01-24
add dynamic import component on Link ViewMode

YeeJiaWei:


pages/api/v1/migration/index.ts
```diff
@@ -7,16 +7,6 @@ import verifyUser from "@/lib/api/verifyUser";
 import importFromWallabag from "@/lib/api/controllers/migration/importFromWallabag";
 import importFromOmnivore from "@/lib/api/controllers/migration/importFromOmnivore";
 
-export const config = {
```
daniel31x13, line 10: Why is this removed?

daniel31x13: More specifically this:

```ts
export const config = {
   api: {
     bodyParser: {
       sizeLimit: process.env.IMPORT_LIMIT
         ? process.env.IMPORT_LIMIT + "mb"
         : "10mb",
     },
   },
 };
```

PR #1017, opened 2025-01-24
fixed height causing double overflow

YeeJiaWei:


daniel31x13:
Thanks for pointing this out!

Issue #932, opened 2024-12-23
Splitting "DATABASE_URL"


RMI78:
**Is your feature request related to a problem? Please describe.**
I would like to implement Linkwarden behind my stack, the thing is it doesn't comply to regular docker standards my stack use to rely on. Such as a lack of docker entrypoint, and most importantly a confusing environment variable `DATABASE_URL` which is in reality a datasource mixing credentials and URL of the database. Which is IMO considered as a very bad practice because there's little to no protection regarding your DB creds which shouldn't have the same confidentiality level as the DB URL. 

**Describe the solution you'd like**
I am willing to try to provide myself a quick first pull request which would be to split the `DATABASE_URL` environment variable into 2 environment variables `POSTGRES_URL`, `POSTGRES_USER` and taking the advantage of `POSTGRES_PASSWORD` which is already implemented. It could provide a more granular control over Docker Compose, the datasource string will then be parsed in the app. On the long run this could lead to an eventual entrypoint.sh for Docker with reading environment variables from files (or doing this directly from the app which I don't know how), thus, bringing Docker Secrets into the game. 

**Describe alternatives you've considered**
The alternative would be to let things as it is as it is only an improvement into better, Docker-friendly, integration. 

**Additional context**
Opening the issue to see how open the maintainers are with this idea instead of straight up coming with a pull request which would have required me some extra work. If you also have any suggestion over what is said above I am all ears out.

PR #779, opened 2024-09-29
Allow running docker images as non-root user

tribut:
This changes permissions inside the docker image in order to be able to run linkwarden without root permissions.

I decided against creating a special-purpose user as proposed in #701, because this would be a breaking change for existing installs (requires changes to ownership of existing bind-mounts). Instead, I've added configuration to docker-compose.yml that new installs will pick up and allows customization by the user.

Fixes #550
Fixes #655 
Fixes #712

PR #1030, opened 2025-01-31
Better docker integration

RMI78:
TL;DR:
# This PR is mostly about:
- Setting all docker files in a dedicated folder and renaming path everywhere for it to work
- Using an entrypoint bash file just like most of the docker images we find in the industry 
- Giving the choice to either work with `DATABASE_URL` or split the values into `POSTGRES_USER`, `POSTRES_PASSWORD` and `POSTGRES_HOST` 
- Introducing compatibilities with Docker Compose secrets with `_FILE` ending name variables 
- Drops root privileges once started

#  Now the long version:
The initial need was to support Docker compose secrets and split `DATABASE_URL` prior to #932 for more granular control over security, then to do that you need an entrypoint file to process environment variables fed by the docker compose file. Which wasn't the case.
The function taking the variables `POSTGRES_USER`, `POSTRES_PASSWORD` and `POSTGRES_HOST` just reassemble the whole thing into a fully functional DATABASE_URL environment variable within the container, and drops any attempts of doing it if any of the splitted variable is empty or if the `DATABASE_URL` is not empty, which make this little change non-regressive as people still can use `DATABASE_URL`.
So I created an entrypoint bash file with latest functionalities as enumered above. Last bullet point is in conflict with #779 and we can talk about it here because it really depends on how the app work. Apparently you are having issues implementing it fully rootless so I toss my 2 coins but can also remove this if you don't want it.
Rougly this PR gets Linkwarden closer to what most dockerized app do and allow more customization  

# Tests
I tested my PR and everything works locally, tried with Github Action on my fork and had an issue with a library when building (during the container releasing flow) which I don't really understand how related is it to my work, enlighten me if you have an idea.
Here are the docker compose files and the env variables I tried my modifications with.
In the first time to see if droping privileges+giving up on `DATABASE_URL` for the splitted variables would works:
```yml
version: "3.5"
services:
  postgres:
    image: postgres:16-alpine
    env_file: .env
    restart: always
    volumes:
      - ./pgdata:/var/lib/postgresql/data
    networks:
      - linkwarden-network
  linkwarden:
    env_file: .env
    restart: always
    build: 
      context: ..
      dockerfile: docker/Dockerfile
    networks:
      - linkwarden-network
    ports:
      - 3000:3000
    volumes:
      - ./data:/data/data
    depends_on:
      - postgres

networks:
  linkwarden-network:
```
and this was the .env
```
POSTGRES_USER=postgresuser
POSTGRES_PASSWORD=postegrestestpassword
POSTGRES_HOST=postgres
NEXTAUTH_URL=http://localhost:3000/api/v1/auth
NEXTAUTH_SECRET=VERY_SENSITIVE_SECRET
```
Then I commented the first 3 lines of the .env, created a secret folder and written the secrets inside. I then modified the docker compose file so it looks like this 
```yml
version: "3.5"
services:
  postgres:
    image: postgres:16-alpine
    env_file: .env
    environment:
      - POSTGRES_USER_FILE=/run/secrets/pg_user
      - POSTGRES_PASSWORD_FILE=/run/secrets/pg_password
    restart: always
    volumes:
      - ./pgdata:/var/lib/postgresql/data
    networks:
      - linkwarden-network
    secrets:
      - pg_password
      - pg_user
  linkwarden:
    env_file: .env
    environment:
      - POSTGRES_USER_FILE=/run/secrets/pg_user
      - POSTGRES_PASSWORD_FILE=/run/secrets/pg_password
      - POSTGRES_HOST=postgres
    restart: always
    build: 
      context: ..
      dockerfile: docker/Dockerfile
    networks:
      - linkwarden-network
    ports:
      - 3000:3000
    volumes:
      - ./data:/data/data
    depends_on:
      - postgres
    secrets:
      - pg_password
      - pg_user

networks:
  linkwarden-network:

secrets:
  pg_password:
    file: ./secrets/pg_password.txt
  pg_user:
    file: ./secrets/pg_user.txt
```


docker/entrypoint.sh
```diff
@@ -0,0 +1,333 @@
...
+          && -z ${DATABASE_URL+x} ]];then
+        export DATABASE_URL="postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:5432/linkwarden"
+    fi
+}
+
+load_secrets() {
+    # All the environment variables will support a `_FILE` suffix that allows
+    # for setting the environment variable through the Docker Compose secret
+    # pattern.
+    local -a secret_supported_vars=(
```
daniel31x13, line 43: Hey really sorry for the late response, is there a way to not to redefine the environment variables here and just use the ones from the `.env.sample` file?

RMI78: Hey, actually I agree with you, this is very dirty and not dynamic. But unfortunately, there is actually no straightforward way to do this as the .env.sample is out of the image. 

While rereading my code, I had a hacky idea. I could actually copy the .env.sample file into the image at build time. Using a combination of sed and other filtering commands, I could parse the file into getting the variables in a list, but that's a quirky way to do things IMO even though it would be "dynamic" and centralized, thus, fitting your needs. Tell me what you think about it and I can totally do that

daniel31x13: > parse the file into getting the variables in a list

That's what I was also thinking of. Like, it'd be great if you could do that instead :)

RMI78: Hi there, it is done.

daniel31x13: I'm trying to build the image locally and it had problem with copying the files. Here's the log:
```
 => ERROR [linkwarden main-app  5/12] COPY ./package.json ./yarn.lock ./playwright.config.ts ./             0.0s
 => CACHED [linkwarden main-app  6/12] RUN --mount=type=cache,sharing=locked,target=/usr/local/share/.cach  0.0s
 => CACHED [linkwarden monolith-builder 2/2] RUN set -eux && cargo install --locked monolith                0.0s
 => CACHED [linkwarden main-app  7/12] COPY --from=monolith-builder /usr/local/cargo/bin/monolith /usr/loc  0.0s
 => CACHED [linkwarden main-app  8/12] RUN set -eux &&     npx playwright install --with-deps chromium &&   0.0s
 => CACHED [linkwarden main-app  9/12] COPY . .                                                             0.0s
 => CACHED [linkwarden main-app 10/12] RUN yarn prisma generate &&     yarn build                           0.0s
 => ERROR [linkwarden main-app 11/12] COPY ./docker/entrypoint.sh /data/entrypoint.sh                       0.0s
------
 > [linkwarden main-app  5/12] COPY ./package.json ./yarn.lock ./playwright.config.ts ./:
------
------
 > [linkwarden main-app 11/12] COPY ./docker/entrypoint.sh /data/entrypoint.sh:
------
failed to solve: failed to compute cache key: failed to calculate checksum of ref : "/docker/entrypoint.sh": not found
```

Are you sure you have it working on your machine?

Also what could go wrong if we remove the following part along with the `load_secrets` function in the entrypoint file?
```
COPY ./docker/.env.sample /data/.env.sample
RUN grep -v "^#" .env.sample \
  | grep -v "^$" \
  | sed 's/=$//' > variables_list.txt && \
    rm /data/.env.sample
```

If it's only to check the postgres connection, we can remove it altogether.

Thanks in advance!

RMI78:
Hello @daniel31x13 , it has been a while, any update regarding that ?

docker/entrypoint.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+# Get PUID/PGID
+PUID=${PUID:-911}
+PGID=${PGID:-911}
+BASH_SOURCE=${BASH_SOURCE:-$0}
+
+add_user() {
```
stavros-k, line 8: this should not happen if id != 0.
I'm running linkwarden using `user: ...` docker compose directive and works just fine

raginggeek:
really hoping to see this PR merge in as I recently moved to docker swarm for my homelab and was looking to repo my compose files, and the lack of secret support limits my ability since I can't use env file interpolation and the recommended path for swarm is to use secrets.

docker/entrypoint.sh
samip5: Wil not work as UID != 0

docker/entrypoint.sh
```diff
@@ -0,0 +1,62 @@
...
+    fi
+}
+
+
+check_postgres_connection(){
+    if    [[ -n ${POSTGRES_HOST} \
+          && -n ${POSTGRES_PASSWORD} \
+          && -n ${POSTGRES_USER}  \
+          && -z ${DATABASE_URL+x} ]];then
+        export DATABASE_URL="postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:5432/linkwarden"
```
samip5: Shouldn't hardcore database port nor database name itself.

PR #1078, opened 2025-03-03
feat(readable): Editor highlighting & comments

il516:


daniel31x13:
Closing this since there is #1093

(Those of you who found out about this, we needed to take another route to do this, and it was my fault for the unclear directions...)

PR #1129, opened 2025-04-11
Fix pagination

Th3On3:
Fix Pagination on getLinks -> returns [response:links: [links], cursor:c, take: t, skip: s] .


Th3On3:
I will pull request later.

PR #1168, opened 2025-05-02
Implement more of the vercel openai sdk bits to allow for api compatible servers

simcop2387:
This should let any other compatible AI servers, proxies, or routers work mostly natively.

I've added the needed variables to the `.env.sample` file along with adding the organization and project fields to the setup so that it can be used for tracking/accounting later on if users want to do that.

I understand that LW is open source but also a paid service, not sure what you need from me as far as licensing goes to include this so I'll leave the following until I'm told what you need otherwise:

1. I don't consider this to be all that significant of work that would merit being copyrightable in the first place, but understand that neither you nor I can really make that choice
2. I am willing to release any claims needed (i.e. CLA or other agreements).

Incidentally I'm about to test this myself with Qwen3-4B locally after it finishes building.  Let me know if there's any changes you'd like me to make since i'm only mildly familiar with typescript so I might not have done something all that idiomatically


simcop2387:
CLA should be signed, and I've confirmed that it's working for me with Qwen3-4B with 32k context at ~100 tokens/sec :) doesn't always decide to create tags but that's all down to the model.  Might be worth some customization of the prompt in the future.

simcop2387:
Set to the proper branch and merged from dev to clear the conflicts.

daniel31x13:
Hey thanks for the PR! Noticed there's an `process.env.OPENAI_NAME` environment variable, is that required to be changed for the third-party providers?

I'll be making some changes to make the `config.compatibility` dynamic based on the `config.baseURL` and merge this afterwards :)

daniel31x13:
Also wouldn't it be better if we remove the `OPENAI_PROJECT` and `OPENAI_ORGANIZATION` environment variables for minimal complexity to set it up? Will that cause any potential issues?

simcop2387:
@daniel31x13 

As far as I could tell what I added for the OPENAI_NAME variable seems to be seems to be something related to the library itself related to compre atibility.  Not entirely sure what it does in the end because the docs aren't all that helpful there.

https://ai-sdk.dev/providers/ai-sdk-providers/openai

Likely setting the compatibility setting dynamically based on if the base is changed is likely to work fine I would think given what it does have there.

The project and organization parameters are entirely optional and I added them while I was there because I know they can be useful for even the official API since they can get used for logging more fine grained billing and other stats for usage if desired.

daniel31x13:
I already used the `@ai-sdk/openai-compatible` package instead which now directly handles everything. Thanks for your help!

PR #1174, opened 2025-05-05
feat(web): add support for http/https proxies

wnor543:


daniel31x13:
Thanks for the contribution!

PR #1175, opened 2025-05-05
add env to disable meilisearch telemetry 

akramer-zibra:
Meilisearch collects telemetry data by default. It is possible to disable meilisearch telemetry with folliwing env MEILI_NO_ANALYTICS=true: https://www.meilisearch.com/docs/learn/self_hosted/configure_meilisearch_at_launch#disable-analytics

This patch adds this env to .env.sample.


daniel31x13:
Hey, thanks for the contribution, I think a better approach would be to add this variable to the [`docker-compose.yml`](https://github.com/linkwarden/linkwarden/blob/dev/docker-compose.yml#L22) instead since using `MEILI_NO_ANALYTICS` won't affect manual installation setups, only docker installations.

akramer-zibra:
Sure, of course. So the idea is to only include necessary configurations in the .env file and put optional ones into the docker-compose.yml? I am fine with that. I was just confused that meilisearch automatically collects telemetry and wanted to point this out briefly. Thanks for your evaluation on this

PR #1183, opened 2025-05-06
fix: http protocol support

redrathnure:
add proper handling of HTTP protocol


daniel31x13:
Hey, this isn't really requested as of now. Thanks for putting your time into it though!

PR #1204, opened 2025-05-29
Meili search only works if key is set

JasonLocklin:
Sets and shares `MEILI_MASTER_KEY` between Linkwarden and Meili search in docker compose file. This will need to be set in the .env file.


daniel31x13:
This is already mentioned in the [docs](https://docs.linkwarden.app/self-hosting/installation).

PR #1215, opened 2025-06-07
Add Synology OIDC as login option based upon Authelia settings successful login

claflico:
This adds Synology SSO OIDC as an option based upon copying code referring to Authelia since I was able to login after using the Authelia env settings.


wlofman021 (approved):

daniel31x13:
Merged! Thanks for the contribution and sorry for getting to this late.

PR #1217, opened 2025-06-08
Fix wrong cases in imports

matejrisek:
Even though the current codebase works on the case-insensitive system, this is unfortunately not a case for case-sensitive systems such as Linux.

This commit fixes cases in imports causing issues for builds on Linux.

### DIFF
--- a/apps/web/components/Announcement.tsx
+++ b/apps/web/components/Announcement.tsx
@@ -1,7 +1,7 @@
 import Link from "next/link";
 import React, { MouseEventHandler } from "react";
 import { Trans } from "next-i18next";
-import { Button } from "@/components/ui/button";
+import { Button } from "@/components/ui/Button";
 
 type Props = {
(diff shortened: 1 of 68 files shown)


daniel31x13:
Hey, there isn't any problem with the installation on linux, and the imports are done properly.

Please open an issue if you had any problems with installation, thanks.

matejrisek:
> Hey, there isn't any problem with the installation on linux, and the imports are done properly.
> 
> Please open an issue if you had any problems with installation, thanks.

Hi, thanks for having a look at this PR.

I can confirm there's no issue due to this being fixed in [another PR](http://github.com/linkwarden/linkwarden/pull/1216), which was created on the same day as this one 😄 

In that PR `Button` file is renamed to the lowercase `button` which effectively resolves all the wrong imports.

daniel31x13:
Oh, got it. Thanks for clarifying!

Issue #907, opened 2024-12-03
Detect CAPTCHAs and browser verification pages, allow recapture after manual intervention


unitof:
**Is your feature request related to a problem? Please describe.**
Many captures are flagged by servers as bots, and redirected to a browser verification or CAPTCHA page. These pages end up being what get snapshotted by Linkwarden, instead of the target page.

**Describe the solution you'd like**
Detect known CAPTCHAs and save them in a queue. Allow human to pass them and then recapture the page they redirect to.

Or, allow capturing via proxy from local device running Linkwarden.

**Describe alternatives you've considered**
Self hosting with fresh IP addresses, or saving and uploading my own HTML snapshots of target pages.

PR #1251, opened 2025-06-27
Support Bypass Cloudflare Captcha using FlareSolverr

SteveImmanuel:
## Summary
This PR integrates support for `FlareSolverr`, a service that solves CAPTCHA challenges on websites protected by Cloudflare. When a page is requested from such sites, `FlareSolverr` will solve the CAPTCHA, retrieve the necessary cookies, and return them. These cookies are then used by the worker's browser instance to ensure the page is archived correctly.

## Problem
Websites like StackOverflow, StackExchange, and others protected by Cloudflare's security measures fail to be archived properly because the CAPTCHA page blocks the content. Without solving this challenge, the archival process cannot proceed as intended.

### Before
When attempting to archive websites protected by Cloudflare, such as StackOverflow or StackExchange, the page fails to load correctly due to the CAPTCHA challenge.
<div style="display: flex;">

</div>

### After
After integrating `FlareSolverr`, the CAPTCHA is solved successfully, and the page is loaded and archived correctly.
<div style="display: flex;">

   
</div>

## Setup
First we need to deploy the `FlareSolverr` service. Then, in the `.env` we can add `FLARESOLVERR_URL` to point to the endpoint of that service, e.g. `FLARESOLVERR_URL=http://localhost:8191/v1`.
- If this key is not set in the `.env`, everything will run the same as before
- If the webpage that is being archived is not protected by CloudFlare, everything should also work normally

Additionally, for websites protected by CloudFlare, the link name is fetched first by the frontend code, resulting in `Just a moment...`. To fix this, when the worker is archiving the page, it will get updated with the correct name according to the title of the website. If during the archival process, the user updated the link name, the final name will remain as what the user provided.

## Additional Notes
- Related issues https://github.com/linkwarden/linkwarden/issues/907
- The captcha solver can be any service as long as the API is compatible with `FlareSolverr`. I found that the original [FlareSolverr](https://github.com/FlareSolverr/FlareSolverr) actually has not been working since last year, and found [FlareBypasser](https://github.com/yoori/flare-bypasser) to be a great alternative as of today (June 27th 2025)


GoogleBot42:
I tried this out with flaresolverr and it is working well for solving non-interactive challenges (I didn't need FlareBypasser).  I have noticed that the title of the webpage isn't getting updated after archival is finished though.

I'm an outsider of the linkwarden project but I'll leave some feedback for things I think should be fixed.

1. Make sure to sign the CLA.
2. Squash your commits into just one
3. It looks like you maybe have inadvertently pulled in some other commits authored by daniel31x13.
4. You have a number of whitespace only changes throughout the file that are unrelated to your changes.  It may help get this merged faster if you have more minimal changes.  I understand that your IDE probably automatically did that but I would override that.

steveimm:
not sure why the CLAssistant has not updated yet, but I've signed the agreement

marcos1:
is there any updates to this?

yurisasc:
It would be nice if Linkwarden can solve Cloudflare challenges. I have the same need because the news provider I follow checks visitors with a Cloudflare challenge.

Thanks for confirming that using Flaresolverr works, @SteveImmanuel. This is exactly what I was looking for. I tried using your changes on Linkwarden's latest release (v2.13.5 at the point of writing), and I found some things to consider:

The Flaresolverr latency on every request is going to impact performance. It's probably better to "lazy-load" the solver, so we only trigger it if we actually hit a challenge page. Also, make sure we're syncing the User-Agent from FlareSolverr back to Linkwarden's Playwright, otherwise the cookies containing `cf_clearance` won't stick.

I also had issues with the article images pulled when using Flaresolverr. For certain news providers, the opengraph image is not shown on Linkwarden after a successful pull. We should probably use `request.get` with a Referer header to get around hotlink blocks on sites like Al Jazeera.

Lastly, Flaresolverr expects request to go through the `/v1/` API. We can normalize the env var to include `/v1` path so we can handle the request whether the user adds the path or not.

I added my own changes on [my fork](https://github.com/yurisasc/linkwarden/releases) and built the docker image. I only built the x86 image because building for ARM takes too long and I only use x86 server personally. I have only tested this on Al Jazeera and tempo.co, but you can test your links by using the custom image or you can hit me up and I'll check them for you.

gthbcrvs:
@daniel31x13 Can this be added? @SteveImmanuel commented that he signed the agreement.

PR #1325, opened 2025-08-06
Fix: request 64px gstatic favicon for consistent icon display

teynar:
Previously, a 32px favicon was requested and scaled to 64px, causing blurriness.

### DIFF
--- a/apps/web/components/LinkViews/LinkComponents/LinkIcon.tsx
+++ b/apps/web/components/LinkViews/LinkComponents/LinkIcon.tsx
@@ -45,7 +45,7 @@ export default function LinkIcon({
       ) : link.type === "url" && url ? (
         <>
           <Image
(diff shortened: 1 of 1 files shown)


daniel31x13:
Thanks for your contribution! Does it fallback to a smaller version if a 64px version wasn't available?

teynar:
> Thanks for your contribution! Does it fallback to a smaller version if a 64px version wasn't available?

If the website has a favicon and a 64px version is not available, Google will upscale it to 64px.

These are the accepted values for the size query param: 16, 32, 64, 128 and 256

If there is a favicon available for the website, Google will provide the icon at the requested size, or an upscaled version in case the website does not have that specific resolution.

If the website does not have a favicon at all, a 16x16 default globe image will be returned.

daniel31x13:
Oh didn't look into it, thanks! 👍

PR #1327, opened 2025-08-06
Fix: the blur on link preview images is too strong

teynar:
The blur effect is very noticeable and can be mistaken as poor image quality.


daniel31x13:
That's actually intentional, users can remove the blur by hiding the link icons.

PR #1364, opened 2025-08-26
Fix jimp unsupported image formats (WebP+avif) by using sharp

spiicytuna:
Fix is to resolve an issue on my docker instances where the worker process crashes when attempting to generate a preview thumbnail for a link using unsupported image formats like AVIF or WebP.

**Problem**

When a link's favicon or preview image is in a format that the `Jimp` library does not support, the worker throws an unhandled error and fails to process the image. This results in crashes (in my case lots of noise in the docker logs) and missing previews in the UI.

This is what I was seeing:

> ```
> Error processing the image: Error: Unsupported MIME type: image/avif
> ```

and

> ```
> Error processing the image: Error: Unsupported MIME type: image/webp
> ```

**Proposed Fix**

A hybrid fallback approach making for it more robust:

  * process as normal with `Jimp`
  * Should `Jimp` fail with `Unsupported MIME type` > catch the error
  * Image buffer pivots to the `sharp` library for conversion into a compatible PNG format
  * Buffer with PNG buffer > `Jimp` for it to continue processing

Originally my compiles were returning server-only `sharp` errors so to prevent inclusion in the frontend browser bundle, duplicated the `generatePreview` utility into server only locations for both `web` and `worker` and updated Next.js config for handling `sharp` as an external mod.

Fix makes sure that Linkwarden can gracefully handle modern image formats without crashing or spitting errors to console.

SharpProcess


daniel31x13:
Hey thanks for your contribution! Just had a brief look at the code, how is `apps/web/lib/server/generatePreview.ts` different from `apps/worker/lib/generatePreview.ts`? Why not merge them?

skeith:
When will this fix be merged?

mainlink0435:
really need this issue resolved. essentially it's broken linkwarden for me.

bwims:
I'm running 2.13.5 , it's March and linkwarden is still unusable. I don't have the resources to run anything but screenshot, and most of my links fail because of the limitations.

Issue #1148, opened 2025-04-18
delete tags with insufficient number of links


ckuethe:
**Is your feature request related to a problem? Please describe.**
After AI tagging runs I have over 3500 tags for my 2700 links and the vast major of the tags only have a single associated link. some even have 0 links (not sure how that happened) 

**Describe the solution you'd like**
I'd like a tool to delete rare tags

**Describe alternatives you've considered**
manually deleting tags. this is also extra annoying because the tag pane resets its scroll position 

**Additional context**

joelis10:
+1, I'd also like something to mass delete tags

spiicytuna:
I ended up wanting the same thing after tagging all of my links and ending up with 178 tags with 0 links.
This works if you know how to manually build, or perhaps will be included in a future release if the Linkwarden team decides to go forward with it:  https://github.com/linkwarden/linkwarden/pull/1376

PR #1376, opened 2025-08-28
Add admin feature for batch cleanup of unused/unwanted tags (by count)

spiicytuna:
Reference:  https://github.com/linkwarden/linkwarden/issues/1148

With the addition of ai tags, depending on the set configuration, many tags show up with only 1 link which either you manually move to another tag or you just swear under your breath wishing they would magically go away.  Beyond that removing links, or collection of links, can leave behind tags with 0 links associated.  Since the tags sit so prominently on the left side having unwanted tags adds clutter.  Deleting 1-2 tags is no issue but in my case there were more than 175 that I wanted gone.

Per a couple of requests I saw of others wanting the same thing, this pull request introduces a new administrative tool to remove unused/unwanted tags based on count.  By default the GUI deletes tags with 0 links associated.  However, adding a new env var "MIN_LINKS_PER_TAG=" to a docker compose file or .env file, the user can set the minimum links to tags that works for their comfort level.  In issue 1148 it sounded like 1-2 links per tag were a preference for that user, while for me it would likely always be tags with 0 links -- but hey this gives options to people who want to configure it to what works for them.

### Testing suggestion

    For testing tags with 1 or more links associated, add the MIN_LINKS_PER_TAG environment variable in 
    the .env file or docker-compose otherwise only tags with 0 links associated will be targeted

    Ensure there are some tags that fall below the threshold

    Login as the admin user (the first account created)

    Go to Settings > Worker

    Verify that the UI correctly displays the threshold matching the env var 
    "Delete all tags with XX or fewer links, applies to all users."

    Clicking the "Confirm" button which will provide the "are you sure" prompt

    Volia! A popup will confirm xx links have been deleted and jumping back to the dashboard 
    the targeted tags should no longer show
    
### Glamour shots

Screenshot From 2025-08-28 13-29-56

Screenshot From 2025-08-28 13-30-13

Screenshot From 2025-08-28 13-30-22

### DIFF
--- a/apps/web/lib/api/controllers/tags/cleanupTags.ts
+++ b/apps/web/lib/api/controllers/tags/cleanupTags.ts
@@ -1,0 +1,21 @@
+import { PrismaClient } from "@linkwarden/prisma/client";
+
+export default async function cleanupTags(minLinksPerTag: number) {
+  const prisma = new PrismaClient();
+
+  const deletedCount = await prisma.$executeRawUnsafe(
+    `
+    DELETE FROM "Tag"
+    WHERE id IN (
+      SELECT T.id
+      FROM "Tag" T
+      LEFT JOIN "_LinkToTag" LTT ON T.id = LTT."B"
+      GROUP BY T.id
+      HAVING COUNT(LTT."A") <= $1
+    );
+  `,
+    minLinksPerTag
+  );
+
+  return deletedCount;
+}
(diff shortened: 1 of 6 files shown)


daniel31x13:
Thanks for your time! In v2.13 you can sort by tag count and bulk delete tags, that should cover the need for this FR.

Issue #530, opened 2024-03-18
Add a button to copy link to clipboard button


savitojs:
**Is your feature request related to a problem? Please describe.**

Working with different browsers (work and personal) and one set to default out of them, It would be nice to have a button to quickly copy the link and paste it to another browser other than the default.

**Describe the solution you'd like**
**Option 1**: It would be nice to add a quick copy link to the clipboard button in the tiles for links (**First screenshot**)
**Option 2**: Add a click link should do action (**Second screenshot**)

**Describe alternatives you've considered**
There are no straightforward alternatives. Now, I click on the tile to open in the default browser copy url and then paste it to another browser.

**Additional context**
![Screenshot from 2024-03-18 10-34-12](https://github.com/linkwarden/linkwarden/assets/5525641/cddfc322-15f4-4a51-8c13-48308cc92eb8)

Or maybe an action here:

![image](https://github.com/linkwarden/linkwarden/assets/5525641/a7fc8608-362e-429d-ab0a-7d8b5ef8a3c9)

spiicytuna:
I am not sure the linkwarden team will go for a UI change that applies to all users but likely is used by a small percentage.  As a compromise I think this is the best bet that the linkwarden team might be able to get behind.

Settings > Preference > Link Settings > Copy link to clipboard

When on clicking any link, it will copy it to the clipboard (for ALL links..), you won't be able to cherry-pick with an icon to copy this link over another you want to open in the existing browser.

_That said if the linkwarden team is OK with a copy-icon for **all** link cards..._  I can go back and add one as a new pull request.  If adding an icon for all link cards... I would think the correct UI behavior would be hiding the icon if Settings > Preference > Link Settings > Copy link to clipboard was checked being with that tuned on, ALL link cards when clicked would copy to clipboard.

PR #1381, opened 2025-08-29
Add link setting for 'Copy link to clipboard'

spiicytuna:
Reference: https://github.com/linkwarden/linkwarden/issues/530

New preference in settings allowing clicking on a link and copying the URL to the clipboard.  In request 530 the use case was users who use multiple browsers whether it be for work or other reasons.

I wasn't sure an icon on each link card would be accepted when this feature is likely not that exciting to most so following existing form and adhering to an uncluttered interface it was added as a global preference.

Settings > Preference > Link Settings > Copy link to clipboard

Code is centralized to a openLink.ts helper func
All other link actions have been tested on a couple docker instances on different hardware and remain working as before

### Glamour shots
Screenshot From 2025-08-29 10-45-05

toast notification:
Screenshot From 2025-08-29 10-45-30

### DIFF
--- a/apps/web/lib/client/openLink.ts
+++ b/apps/web/lib/client/openLink.ts
@@ -4,6 +4,7 @@ import {
 } from "@linkwarden/types";
 import { generateLinkHref } from "@linkwarden/lib/generateLinkHref";
 import { LinksRouteTo } from "@linkwarden/prisma/client";
+import { toast } from "react-hot-toast";
 
 const openLink = (
   link: LinkIncludingShortenedCollectionAndTags,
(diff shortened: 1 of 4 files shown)


daniel31x13:
Why not right click on the link url and just copy the link that way?

image

spiicytuna:
@daniel31x13 - In Windows Firefox, Linux Firefox, Linux Edge, OSX Safari -- using your latest release as-is not anything I have modified I do not have the option to copy the link (screenshot below Windows Firefox but looks identical in Firefox Linux)

image

spiicytuna:
@daniel31x13 - (more testing) if I right-click on the folder the link belongs to I can get that to let me copy the link....which leads me to believe it's not one of my extensions blocking the option (although I cannot rule that out w/o disabling them all).

image

daniel31x13:
@spiicytuna oh you need to right click on the url itself

Issue #702, opened 2024-08-14
[BUG] Copy to Clipboard button not working


deffcolony:
**Describe the bug**
The Copy to Clipboard button does not copy the access-token into clipboard the console says:

`Uncaught TypeError: navigator.clipboard is undefined`

`access-tokens-be43b94f2e616854.js:1:11674 /_next/static/chunks/pages/settings/access-tokens-be43b94f2e616854.js`

**To Reproduce**
1. Navigate to: settings --> access-tokens
2. Click on New Access Token
3. Click Copy to Clipboard (which does not copy to clipboard)

**Expected behavior**
That the Copy to Clipboard copies the access-token to clipboard

**Screenshots**

https://github.com/user-attachments/assets/15f4ab0d-8460-43c9-b0f2-d4b18de0801b

![image](https://github.com/user-attachments/assets/92f2d5be-d695-4e24-ad96-8f147d0d9213)

**Desktop (please complete the following information):**
 - Docker
 - Floorp browser 11.16.0 (64-bit)
 - Linkwarden v2.6.2

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

daniel31x13:
You need to host Linkwarden on HTTPS instead of HTTP, this is because the [`navigator.clipboard`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) property needs to be executed over a secure context.

PR #1436, opened 2025-10-03
CopyText hook and fallback secure context

marcelotk15:
ref #702


daniel31x13:
Hey thanks for your time! I'm thinking that most people would _(should)_ already be using HTTPS anyways, how will this PR be beneficial?

marcelotk15:
Good question — and you’re right that most production apps run over HTTPS nowadays.
This change doesn’t affect them at all: when the connection is secure, the hook still uses the standard `navigator.clipboard.writeText()` path exactly as before.

The only difference is that we now gracefully **fallback** when the environment isn’t considered secure by the browser.
That covers a few common dev/test scenarios:

* Running locally via **raw IP** (`http://127.0.0.1`, `http://192.168.x.x`) or over LAN.
* Using a **reverse proxy on a local network** with a fake TLD (e.g. `myapp.local`) **without real TLS certificates**.

Even though `localhost` itself is treated as secure, **anything accessed by IP or fake local domains without TLS is not**, and that silently breaks the Clipboard API.
The fallback ensures “copy to clipboard” continues working seamlessly in those cases.

If it makes sense, we could also expose a small config option to bypass the HTTPS check (e.g. `preferFallbackInInsecureContext`), so users using local proxies or LAN domains can force the fallback without changing production behavior.

daniel31x13:
Sorry, but this isn’t really needed. It could cause too many bugs with all the edge cases, and it’s just not worth it.

Issue #1392, opened 2025-09-06
Feature Request: Generate Formats From HTML Archive


JadoJodo:
**Is your feature request related to a problem? Please describe.**
While using the SingleFile upload results in very clean, readable snapshots, the other formats will often be broken or incomplete due to ads, paywalls, etc. (reasons for which the SingleFile local snapshot option was used). This is particularly prevalent when reading an article generated from SingleFile results in notes and/or highlights, but the Readable archive is broken (paywall, incomplete page).

**Describe the solution you'd like**
Add a setting (or alternate endpoint for SingleFile) that will cause the PDF, Readable, and Screenshot archives to be generated _from_ the uploaded web archive. I suspect this would either involve navigating Playwright to the uploaded HTML bundle in a local, tmp directory OR (perhaps) LinkWarden hoisting the file publicly somewhere and generating the other views that way (privacy issues here, though). 

**Describe alternatives you've considered**
The ability to annotate the HTML version of the archive directly.

**Additional context**
I find this issue occurs quite frequently with Substack and premium news sites.

JadoJodo:
🤦‍♂️ Looks like most of this may be covered in #1387

daniel31x13:
Fixed in [v2.13.1](https://github.com/linkwarden/linkwarden/releases/tag/v2.13.1) (thanks to @supercoolspy)

Issue #1386, opened 2025-09-02
Singlefile is usable, but still has many flaws.


mapleshadow:
HTML pages saved with Singlefile have no title and no AI-generated tags. The saved page only contains HTML, with no PDF or visual view page.

daniel31x13:
Fixed in [v2.13.1](https://github.com/linkwarden/linkwarden/releases/tag/v2.13.1) (thanks to @supercoolspy)

PR #1448, opened 2025-10-11
fix: support using monolith content for other archive formats

### LINKED ISSUE #1384: New "Upload from SingleFile" feature and Readable format
Hi. I tried the new SingleFile feature to preserve a document from "htmlpreview.github.io/?[URL]" because none of the preservation formats work for links using this service. With SingleFile i get the HTML format stored correctly but still not the Readable format. So i was wondering whether the Readable could be generated from the HTML when creating a link using SingleFile?

supercoolspy:
Fixes #1384 #1392 and #1386


daniel31x13:
Thank you for your contribution!

Issue #712, opened 2024-08-17
2.6.0+ no preserved formats


mdlaat:
**Describe the bug**
Probably since updating to 2.6.0 my selfhosted (in docker) linkwarden seems unable to create preserved formats.
When trying to display the preserved formats, I see a message: "Link preservation is in the queue"
I see this with links added from july 22 and more recent.

Links added from july 17 and earlier are ok.

**To Reproduce**
1. Add a link
2. Click on '...' and 'preserved formats' on that link
3. See error

**Desktop (please complete the following information):**
 - OS: ubuntu 22.04 
 - selfhosted docker

**Errors from docker logs**
warning Skipping preferred cache folder "/.cache/yarn" because it is not writable.
yarn run v1.22.19
warning Selected the next writable cache folder in the list, will be "/tmp/.yarn-cache-1007".
warning Cannot find a suitable global folder. Tried these: "/usr/local, /.yarn"
warning Skipping preferred cache folder "/.cache/yarn" because it is not writable.
warning Selected the next writable cache folder in the list, will be "/tmp/.yarn-cache-1007".
$ /data/node_modules/.bin/prisma migrate deploy
Prisma schema loaded from prisma/schema.prisma
Datasource "db": PostgreSQL database "postgres", schema "public" at "postgres:5432"

42 migrations found in prisma/migrations

No pending migrations to apply.
Done in 0.93s.
yarn run v1.22.19
warning Cannot find a suitable global folder. Tried these: "/usr/local, /.yarn"
$ concurrently -P "next start {@}" "yarn worker:prod" --
[0] - ready started server on 0.0.0.0:3000, url: http://localhost:3000
[1] warning Skipping preferred cache folder "/.cache/yarn" because it is not writable.
[1] warning Selected the next writable cache folder in the list, will be "/tmp/.yarn-cache-1007".
[1] $ ts-node --transpile-only --skip-project scripts/worker.ts
[1] warning Cannot find a suitable global folder. Tried these: "/usr/local, /.yarn"
[1] Starting the link processing task

 
**and**

[0] Failed to write image to cache dMi6oj+-UkXSAlE4CvAmbi-A4J4XwXNxDC1NFuHw6dM= [Error: EACCES: permission denied, mkdir '/data/.next/cache/images'] {
[0]   errno: -13,
[0]   code: 'EACCES',
[0]   syscall: 'mkdir',
[0]   path: '/data/.next/cache/images'
[0] }

[...]

Issue #655, opened 2024-07-19
After updating to 2.6.0, Preserved Formats seems to be unable to generate.


dragonjay-lyj:
**Describe the bug**
In the previous version (was it 2.6.0?), at least one Readable format could be generated, but in this latest version, nothing is produced.

**To Reproduce**
Normal steps for adding a link

**Expected behavior**
At least one Readable format can be generated.

**Screenshots**
![mmexport1721423389602.png](https://github.com/user-attachments/assets/7aeb3dfe-029e-407d-a160-24ce9183790e)

**Desktop (please complete the following information):**
 - OS: [Windows 11]
 - Browser [chrome]
 - Version []

**Smartphone (please complete the following information):**
 - Device: [Redmi K60 pro]
 - OS: [Android 14]
 - Browser [chrome]
 - Version []

**Additional context**
I took a look at the log (as shown in the image) and it seems that the generation failure is due to a timeout error. Is it possible to extend the time limit?

daniel31x13:
How about other links? Same issue? (Like [this](https://news.ycombinator.com) one)
Also did you take the manual installation or installation using docker compose?

dragonjay-lyj:
> How about other links? Same issue? (Like [this](https://news.ycombinator.com) one) Also did you take the manual installation or installation using docker compose?

Ah, when I first installed it, you hadn't updated the version yet, which was 2.6.0. At that time, I installed it using Docker and was able to generate one Readable format. Later, I wanted to try a manual installation, so I deleted Docker, but it was around this time that you updated the version. To test my theory, I tried both Docker and manual installation in the new version, and neither method was able to generate a Readable format.

dragonjay-lyj:
> How about other links? Same issue? (Like [this](https://news.ycombinator.com) one) Also did you take the manual installation or installation using docker compose?

Regarding the link, as I mentioned, in the previous version, it was possible to generate a Readable format.

dragonjay-lyj:
Also, forgive me for getting the version number wrong. The latest version should be 2.6.0.

dragonjay-lyj:
I’ve confirmed it. I just tested and it is indeed an issue with the new version. I rolled back to 2.5.2 and successfully generated a Readable format.

<img width="744" alt="c49c7e0e70066679911ec4e32eac6b5" src="https://github.com/user-attachments/assets/94b4dded-a665-441a-9d70-e5acd970df63">

I encountered some issues during the rollback process.

[...]

Issue #550, opened 2024-04-02
Cannot create prevesered formats without root


monstermuffin:
**Describe the bug**
When running the Linkwarden container as a non-root user, preserved formats never get created because `npx playwright install` doesn't work as per the logs below.

I do not run my containers as root, my deployment with ansible creates a user per container, sets this user as owner of the container's data directory and runs the container as said user.

**To Reproduce**
Define a user to run the container as using compose using `user`. 

I do this via ansible. `user: "{{ linkwarden_uid }}:{{ linkwarden_gid }}"`

**Expected behavior**
playwright should be installed without requiring root perms.

**Log**
```
[1] Processing link <somelink> for user 1
[1] Error processing link <somelink> for user 1: browserType.launch: Executable doesn't exist at /.cache/ms-playwright/chromium-1067/chrome-linux/chrome
[1] ╔═════════════════════════════════════════════════════════════════════════╗
[1] ║ Looks like Playwright Test or Playwright was just installed or updated. ║
[1] ║ Please run the following command to download new browsers:              ║
[1] ║                                                                         ║
[1] ║     npx playwright install                                              ║
[1] ║                                                                         ║
[1] ║ <3 Playwright Team                                                      ║
[1] ╚═════════════════════════════════════════════════════════════════════════╝
[1]     at archiveHandler (/data/lib/api/archiveHandler.ts:34:34)
[1]     at archiveLink (/data/scripts/worker.ts:98:27)
[1]     at /data/scripts/worker.ts:119:17
[1]     at processBatch (/data/scripts/worker.ts:119:6)
[1]     at async init (/data/scripts/worker.ts:134:7) {
[1]   name: 'Error'
[1] }
```

**Additional context**
If I try to run `npx playwright install` manually it will fail.

```
npx playwright install
npm ERR! code EACCES
npm ERR! syscall mkdir
npm ERR! path /.npm
npm ERR! errno -13
npm ERR!
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR!
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 999:994 "/.npm"

npm ERR! Log files were not written due to an error writing to the directory: /.npm/_logs
npm ERR! You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
I have no name!@ece36c86f950:/data$ Failed to install browsers
[...]

PR #1482, opened 2025-11-08
Allow running Docker container as non-root user

holocronweaver:
This is a modification and rebase of #779 to allow running Docker containers as a non-root user.

Main changes from that PR:

- went with PUID/PGID variable names to match the linuxserver.io convention, which avoids conflicts with reserved UID/GID variables.
- setting `HOME` environment variable to `/home` so Playwright is installed there and can be accessed by non-root users (setting `ENV PLAYWRIGHT_BROWSERS_PATH` did not do the trick for me).
- setting permissions in an entrypoint script (requires root), then dropping root privileges using gosu if PUID/PGID are set.

I verified that the Docker image size did not change between `linkwarden:latest` and this new image (both 2.48GB), which I saw a concern in #779. The reason that PR caused the size increase was due to its `chmod -R ... /data` inside the RUN command - changing permissions in this way creates a new layer which duplicates all the files in `/data` with the new permissions. By switching to an entrypoint script that sets permissions the layer-duplication problem is avoided.

I tested:
- running without a custom user
- running with custom PUID/PGID
- running on linkwarden:latest, then upgrading to new image without changing anything else (to verify backwards compatibility)

In all cases I was able to create an account, login, and add links without any diverging warnings or errors in the logs. The backwards compatibility case had no issues.

Fixes https://github.com/linkwarden/linkwarden/issues/550
Fixes https://github.com/linkwarden/linkwarden/issues/655
Fixes https://github.com/linkwarden/linkwarden/issues/712

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -8,6 +8,13 @@ DATABASE_URL=
 # Docker installation database settings
 POSTGRES_PASSWORD=
 
+# Run processes inside container as specific user/group. Defaults to root.
+# WARNING: if docker is run using `sudo`, create all mounted directories with
+# correct ownership before starting the container, or else the directories will
+# be created with root ownership which can conflict with the uid/gid set here.
+PUID=
+PGID=
+
 # Additional Optional Settings
 PAGINATION_TAKE_COUNT=
 STORAGE_FOLDER=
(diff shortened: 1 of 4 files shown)


holocronweaver:
Closing this to use a branch based on dev: #1483

Issue #369, opened 2023-12-28
Archiver doesn't work in 2.4.1 when using a custom storage path and running as non-root


samip5:
**Describe the bug**
Adding an URL to a collection results in an playwright error due to not finding an executeable as it's trying to find it at /.cache/*.

**To Reproduce**
1. Run Linkwarden without root privileges aka non-privileged
2. Add link to an collection
3. See the error in processing that said link

**Expected behavior**
I would have expected it to find the chrome executable.

**Screenshots**
![Screenshot of log](https://z.skylab.fi/i/2023-12-28_14-30-21.png)

daniel31x13:
> 1. Run Linkwarden without root privileges aka non-privileged

Well does it work in root privileges? I remember in another issue a user pointing this out as well and the problem was for not running as root.

samip5:
> the problem was for not running as root.

Sorry, but that's not the fix and I'm not conformable running it as root. The issue is that node should be instructed to use the storage path which appears to be using it's default which will not work even if running as root if rootFilesystem is read-only.

daniel31x13:
This is what exactly happens when the app creates a file, feel free to check it out:
https://github.com/linkwarden/linkwarden/blob/main/lib/api/storage/createFile.ts

samip5:
> This is what exactly happens when the app creates a file, feel free to check it out: https://github.com/linkwarden/linkwarden/blob/main/lib/api/storage/createFile.ts

I'm in the Discord. https://discord.com/channels/1117993124669702164/1123628004145369149/1189905698025373716

samip5:
After setting `HOME` env variable to my storage path, and running `npx playwright install` it was able to install them and then it started to work even without root. This sounds like the container building might need some additional things..

daniel31x13:
Ok sounds good.
Gonna change it later... Thanks for pointing it out.

discotimetraveler:
I am having this same issue on my NAS. I did not install using the standard process and am running as non-root. But in my attempts to resolve this I'm a bit confused. Since this is on my NAS I'm not sure where/how I would run `npx playwright install` (not looking to install dev tools on my NAS, and this should  already be in the container?). This is the error I get...

Executable doesn't exist at /.cache/ms-playwright/chromium-1067/chrome-linux/chrome

When I log into the container and poke around for chrome I find...

./data/node_modules/playwright-core/lib/server/chromium

Which is not even close to the path in the error.  

[...]

PR #1483, opened 2025-11-08
Allow running Docker container as non-root user

holocronweaver:
Edit: Updated approach to use an additional final build stage that adjusts `/data` permissions at build time rather than runtime, still without any image bloat (actually shrinks the image by 10MB).

This is a overhaul and rebase of #779 to allow running Docker containers as a non-root user.

Main changes from that PR:

- went with PUID/PGID variable names to match the linuxserver.io convention, which avoids conflicts with reserved UID/GID variables.
- setting `HOME` environment variable to `/data/home` so Playwright is installed there and can be accessed by non-root users (setting `ENV PLAYWRIGHT_BROWSERS_PATH` did not do the trick for me).
- introduce a new final stage that combines the results of the monolith-build and app-build stages, then fixing permissions to allow global write on /data without duplicating layers in the final image.

Upside is now the image allows the standard --user approach to specifying
UID/GID in Docker Compose, simplifying setup for users.

I verified that the Docker image size did not incrcease between `linkwarden:latest` and this new image (latest: 2.48GB, this: 2.47GB, so 10MB savings), which I saw a concern in #779. The reason that PR caused the size increase was due to its `chmod -R ... /data` inside the RUN command - changing permissions in this way creates a new layer which duplicates all the files in `/data` with the new permissions. By adding a final layer this avoids adding image bloat when changing permissions since the app-build layer is discarded (actually strinks image by 10MB), though roughly doubles space required to build image (~2GB -> ~4GB) and increases build time by a few seconds. This seems like a small price for the security benefits of enabling to run as non-root user.

## Testing
If anyone wants to take this for a spin, I uploaded an image to `holocronweaver/linkwarden:customeruserV2` built from a cherrypicked `main` branch.

I tested:
- running without a custom user
- running with custom PUID/PGID
- running on linkwarden:latest, then upgrading to new image without changing anything else (to verify backwards compatibility)
- running on linkwarden:latest with custom user workarounds like [this](https://github.com/linkwarden/linkwarden/issues/369#issuecomment-3504628658), then upgrade to new image without changing anything else (to verify common workarounds are not broken)

In all cases I was able to create an account, login, and add links without any diverging warnings or errors in the logs. The backwards compatibility case had no issues.

## Migration

Those running Linkwarden as root or using previous non-root user workarounds need no changes.

Fixes https://github.com/linkwarden/linkwarden/issues/550
Fixes https://github.com/linkwarden/linkwarden/issues/655
Fixes https://github.com/linkwarden/linkwarden/issues/712


holocronweaver:
Updated PR to add a final Docker build stage that adjusts `/data` permissions at build time rather than runtime, still without any image bloat (actually shrinks the image by 10MB).

This enables Docker Compose `--user` field to be user with Linkwarden service, as @MrModest suggested, and avoids any need for those using most custom non-root user workarounds to migrate.

holocronweaver:
Docker image for the new approach, if anyone wants to test and give feedback: `holocronweaver/linkwarden:customeruserV2`

As usual, I recommend not testing this on production data, at least not without backups.

monstermuffin:
Hey @holocronweaver - thanks for this! I was about to ask for https://github.com/linkwarden/linkwarden/issues/550 to be reopened and saw you’re already on it.

I’ve been testing a similar fix and did a few things that I think work better:

- Bake Playwright browsers (and sharp optionally) during build. This removes runtime downloads, avoids `/.cache`/`.npm` permission issues for non-root, speeds cold starts, and shrinks the runtime attack surface.
- Set `HOME` to a writable path and pre-create the Next.js cache (`/data/apps/web/.next/cache`) as writable. Instead of `777`, use group-writable with `chgrp 0` + `chmod g+rwX` so `--group-add 0` / `fsGroup: 0` works without making everything world-writable.
- Add a tiny optional entrypoint: if PUID/PGID are provided, create the user/group, chown known write paths (`/data/apps`, `/data/data`, `/ms-playwright`), then `gosu` drops privileges. If unset, it runs as root (backward-compatible). This avoids `--user` quirks while still supporting arbitrary UIDs.

Happy to share my version if useful. Either way, great to see some movement on this as this was the reason I never tried using LW properly previously.

oxivanisher:
Is this PR still considered? With all the CVE running wild, I wanted to increase the security and realized, that linkwarden can not (yet?) run as non-privileged user.

vnghia:
Actually you can already run with non-root user if you spawn another docker container for chronium `ghcr.io/browserless/chromium:v2.48.3` (the latest compatible version) and set `PLAYWRIGHT_WS_URL` as long as supply the `--user 1000:1000`. I'm using that myself.

To be honest I think we should just document this approach instead of messing around with node and permission because it could break a lot more thing.

Issue #788, opened 2024-10-06
Change timeout for SSO?


zodac:
**Is your question related to a problem or code? Please describe.**
Is it possible to increase the timeout for SSO requests? My Authentik instance is on another server an there is some times a delay between devices. I'd like to be able to set an environment variable or something to extend this.

**Describe what you've tried to solve this question**
Just retry when the connection is better.

**Include any code or screenshots (if applicable)**
I'm basing this off the error message in the logs when my authentication request to Authentik fails:

`message: 'outgoing request timed out after 3500ms'`

AnonymousX86:
I'm facing the same issue. Full error:

```
[next-auth][error][SIGNIN_OAUTH_ERROR]
https://next-auth.js.org/errors#signin_oauth_error outgoing request timed out after 3500ms {
  error: {
    message: 'outgoing request timed out after 3500ms',
    stack: 'RPError: outgoing request timed out after 3500ms\n' +
      '    at /data/node_modules/openid-client/lib/helpers/request.js:137:13\n' +
      '    at async Issuer.discover (/data/node_modules/openid-client/lib/issuer.js:144:24)\n' +
      '    at async openidClient (/data/node_modules/next-auth/core/lib/oauth/client.js:16:14)\n' +
      '    at async getAuthorizationUrl (/data/node_modules/next-auth/core/lib/oauth/authorization-url.js:70:18)\n' +
      '    at async Object.signin (/data/node_modules/next-auth/core/routes/signin.js:38:24)\n' +
      '    at async AuthHandler (/data/node_modules/next-auth/core/index.js:260:26)\n' +
      '    at async NextAuthApiHandler (/data/node_modules/next-auth/next/index.js:22:19)\n' +
      '    at async auth (/data/.next/server/pages/api/v1/auth/[...nextauth].js:1585:12)',
    name: 'RPError'
  },
  providerId: 'authentik',
  message: 'outgoing request timed out after 3500ms'
}
```

Which is weird because I have set up eight OIDC providers in Authentik (alongside other types, but that's not important, I guess), daily used by 2~4 people, and I've never came across timeout issue.

My config for context.

```env
# .env
# Required fields
NEXTAUTH_URL="https://linkwarden.example.com/api/v1/auth"
NEXTAUTH_SECRET="REDACTED"
POSTGRES_PASSWORD="REDACTED"
# General login
NEXT_PUBLIC_DISABLE_REGISTRATION="false"
[...]

PR #1487, opened 2025-11-09
Add configurable HTTP timeout for Authentik provider

krim404:
Suggestion as a fix for https://github.com/linkwarden/linkwarden/issues/788
Edited the function to implement the same function as for the google authentication. New env: AUTHENTIK_HTTP_TIMEOUT defaults to 10000 (10s)


krim404:
wrong pull target, reopened with corrent one (dev instead of main) -> https://github.com/linkwarden/linkwarden/pull/1488

Issue #1470, opened 2025-10-23
Auto delete tags with no links attached


lethallybad:
i'm using ai-tagging and have a few bookmarks. as i've added and deleted a lot of bookmarks i have dead tags hanging around with no links attached. i think it would be cool to add an option in the settings to remove a tag if the last link using it gets deleted. or add a sweep button to delete any tags that aren't attached to any links. sometimes ai will create 4 new tags for the bookmark, i delete bookmark, then i have 4 tags linking to nothing

opazenha:
@daniel31x13 Hello there!
I am willing to help here since I love using Linkwarden.
I am new to development and open source in general.
I believe this is something I can try to assist.
May I submit a PR for this?
Thank you!

PR #1506, opened 2025-12-03
PR: Auto-clean orphan AI-generated tags when links are removed - #1470

opazenha:
Summary
Single-link deletions now clean up AI tags – When a link is deleted, any AI-generated tags that no longer belong to another link are removed immediately, preventing “ghost” tags from piling up in users’ collections. 
Bulk deletions receive the same treatment – Deleting several links in one request now batch-removes orphaned AI tags for all affected owners.
New tag-maintenance helpers – Added APIs to remove all orphaned AI tags for a user, or only those tied to a specific link, so we can expand UI controls later without reworking backend logic.

Testing
Deleted a single link that had unique AI tags - tags disappeared as expected.
Deleted multiple links sharing tags - tags only removed when no remaining links referenced them.
Tried deleting links with no AI tags - no tag deletions triggered, confirming guard rails work.

Considerations for Settings UI
“Sweep orphan tags” button – The new removeAllOrphanTags controller gives us everything needed for a one-click cleanup in Settings. A button could call this endpoint and show how many tags were removed (or if nothing was found). I noticed there are responses with descriptive text to them, so I kept it.
Optional auto-clean toggle – Some users might prefer manual control, so exposing a toggle like “Auto-delete AI tags when their last link is removed” keeps flexibility. The current backend already deletes orphans during link removal, but the toggle could decide whether we invoke that cleanup or skip it for advanced workflows. A user preference flag stored alongside other settings would do the trick, and the existing helper gives us the required logic without additional queries.

Let me know of your feedback! Thanks!

### DIFF
--- a/apps/web/lib/api/controllers/links/bulk/deleteLinksById.ts
+++ b/apps/web/lib/api/controllers/links/bulk/deleteLinksById.ts
@@ -1,8 +1,8 @@
-import { prisma } from "@linkwarden/prisma";
-import { UsersAndCollections } from "@linkwarden/prisma/client";
 import getPermission from "@/lib/api/getPermission";
 import { removeFiles } from "@linkwarden/filesystem";
 import { meiliClient } from "@linkwarden/lib";
+import { prisma } from "@linkwarden/prisma";
+import { UsersAndCollections } from "@linkwarden/prisma/client";
 
 export default async function deleteLinksById(
   userId: number,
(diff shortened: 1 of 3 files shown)


daniel31x13:
Hey, thanks a lot for the PR! I really appreciate the effort you put into the cleanup logic and the consideration for future UI hooks.

That said, I’m going to close this PR for now. While the current behaviour removes orphaned tags, it could still be confusing from a user perspective. For example, if a user has only a handful of tags and bulk-deletes links, seeing several tags disappear automatically might feel unexpected or even data-lossy, particularly if they were relying on those tags conceptually rather than on active links.

Right now, we already offer a more explicit and user-controlled way to handle this: users can bulk-remove unused tags by sorting the Tags page by "Link count" and deleting from there. This keeps the action intentional and visible.

image

Looking ahead, we might explore something like a “Remove unused tags” action directly on the Tags page, but that’s more of a future consideration rather than something we’re planning to take on right now.

Thanks again for taking the time to contribute!

Issue #968, opened 2025-01-07
Ability to share collections INCLUDING all sub-collections


tom666-debug:
**Is your feature request related to a problem? Please describe.**
Can we please have the functionality to share a collection including all sub-collections. 
I have seen some similar requests here but there where not even answered. e.g. https://github.com/linkwarden/linkwarden/issues/628

In our use-case we have for instance a "main" collection holding 80 sub-collections. Instead of for just only the main collection  we must maintain permissions for 81 collections in total. From my perspective this affort could be avoided.

**Describe the solution you'd like**
Something like it was alredy described here (feature 1): https://github.com/linkwarden/linkwarden/issues/628

Danie10:
I second this. I have a main public share with two sub-collections which are also publicly shared. The view I see as a logged in user shows the sub-collections (as shown in image below) but I'd like my public share users to also see that same view. All they see is the main share and I have to put links in the description to the sub-collections.

Maybe this is being worked on, as I don't recall seeing those sub-collections in my view before?

![Image](https://github.com/user-attachments/assets/90a1eeda-198d-47bf-85e0-02568b8d8e36)

Danie10:
Is there any plan to actually implement this? It has been requested by quite a few people, in more than one logged issue.

daniel31x13:
@Danie10 Yep it's planned! We even have an open PR (thanks to @9helix).

Danie10:
@daniel31x13 thanks that is fantastic news - the cherry on the top!

jcolinger3:
Any progress updates?

tom666-debug:
> Any progress updates?

My request (see my first post) was fully integrated with the latest release, works like charm now.

Danie10:
I'm wondering if it only works for logged in members? I ticked the option on, but don't see the sub-collections on my public share. Not sure maybe if there is a public role I should add?

<img width="580" height="272" alt="Image" src="https://github.com/user-attachments/assets/64fbbef8-cb50-49ca-9258-1141cf7049d7" />

tom666-debug:
You are right, for me it also works for logged in members only.

Issue #628, opened 2024-06-28
Features Suggestions - Herited sharing, owner edition...


nothing-fr:
**Add an option for inheriting share permissions**
Is it possible to add an option to choose whether to propagate user rights to sub-collections ?

**Possible solution**
Add a checkbox, Here's an idea:
![image](https://github.com/linkwarden/linkwarden/assets/92157590/dbb9b501-9695-403f-a7ef-9d01003a6dbf)

---

**Possibility of changing the owner of a collection**
I know that an administration page is currently under development (https://github.com/linkwarden/linkwarden/issues/217).  Are there any plans to manage collections with the possibility of changing the owner or transferring it to another person ?

**Possible solution**
Make it possible for an administrator to manage collections via the future Linkwarden administration page

bluenevus:
+1 to feature 2

PR #1568, opened 2026-01-02
Ability to propagate changes in the collection's permissions to all subcollections

9helix:
Upon opening the Share and Collaborate modal of a collection, the user has the ability to tick the checkbox that will propagate members and their permissions for the current collection to all of its subcollections.
Solves #968 and feature 1 from #628.


daniel31x13:
Thanks for your work!

Issue #1633, opened 2026-03-17
Preserved links  for uploaded images are not available in public lists


karabays:
**Describe the bug**
When a PDF or image is added to a public list, linkwarden succesfully creates the preserved formats, but the links for those formats are not available on the public list. when clicked list, it tries to open non-public link and asks login.

**To Reproduce**
1. Create a public list
2. upload a pdf or image
3. open the public list link in incognito tab
4. click the link
5. linkwarden tries to open `https://domain.name/preserved/9?format=2` instead of `https://domain.name/**public**/preserved/9?format=2`

**Expected behavior**
if the list is public, it should open the public version of the preserved format

**Screenshots**

<img width="640" height="376" alt="Image" src="https://github.com/user-attachments/assets/45acb565-7629-49ce-936b-4d56eb8b36a4" />

**Desktop (please complete the following information):**
 - OS: windows
 - Browser: edge
 - Version: 2.13.5 selfhosted

PR #1634, opened 2026-03-17
fix: preserved links in public lists should use public URLs

saschabuehrle:
## Bug
[Issue #1633](https://github.com/linkwarden/linkwarden/issues/1633) — When a PDF or image is added to a public list, the preserved format links are not accessible in public view because they point to private URLs instead of public URLs.

## Fix
Fixed the boolean logic in `PreserverdFormatRow.tsx` where `isPublic` was incorrectly set to `undefined` instead of `false` for non-public pages. This caused the ternary operator to always generate non-public URLs.

## Testing
Verified the fix by tracing through the logic:
- For public pages: `router.pathname.startsWith("/public")` returns `true`
- For private pages: `router.pathname.startsWith("/public")` returns `false` 
- The ternary operator `isPublic ? "/public" : ""` now correctly includes "/public" in the URL for public lists

This ensures preserved format links in public lists use the correct public URL format: `/public/preserved/{id}?format={format}`

Greetings, saschabuehrle


daniel31x13:
Thanks for pointing this out!

Issue #1431, opened 2025-10-01
RAM Usage to high


seals187:
Hello Guys, 

Im trying Linkwarden on pikapods, there i have a VPS with only, "linkwarden" running on that machine, 
with 2 CPU Cores and 2GB of RAM, and the RAM is allready almost 100% used. 

Is that really the case ?
That would be really bad memory management. 

Can you analyse that ? Tell me how to define a max memory treshold. that Linkwarden can use ?

best reegards

saikimecky:
Same problem her while running as a Docker Container. And i cant see why this happens. Its from the start of the container about 1.5G of RAM. 

[Screenshot](https://ibb.co/vvg8GV4J)

seals187:
I dont know how linkwarden scrapes websites, if they are using chrome
like https://changedetection.io/ , I'm running that too on my servers, 
and for changedetection.io, chrome uses ~400-500MB , depends if active
or standby. So if linkwarden uses that to , that would make 400-500MB of RAM used, 
so the REST is 1GB / 1.5GB of RAM used, for xx ??, in my case. 

Other potential candidates ? the search- and db-engine are not
possible  ?, (in your case at least, as like you showed in the printscreens they do not consume more than 1GB together)

maybe some camouflaged mining of Crypto ? =)

ThiaudioTT:
Same here. 

Related issue: #744

seals187:
thx for the update, juste for comparing:

- Readeck does the same (plus,minus)
- 40MB of RAM usage (idle), and possible peaks to 500MB when getting images etc

<img width="680" height="388" alt="Image" src="https://github.com/user-attachments/assets/3567e009-523e-405f-a893-470fe350ff65" />

lirc572:
Not sure what's going on with my new installation that's consuming over 10GB of RAM... I was expecting < 1GB before installation... 

<img width="688" height="264" alt="Image" src="https://github.com/user-attachments/assets/c70f3146-3706-4a80-a8b2-a23e1dbf8b6f" />

seals187:
puuh., damn, thats much.
I had 2 GB on my Container who was configured with 2GB RAM,
and Linkwarden took almost everything of that 2GB of RAM.

So maybe it's a memory leak ? or something like that?

mozarchbtw:
Just setup mine, and its taking approximately 1.4GB on my 4GB system.

ArtCodesArt:
[...]

Issue #1396, opened 2025-09-11
Memory Limits Set on the Container


psarchives:
Was having issues with runaway memory getting to 1.3gb memory usage for the Linkwarden Docker container. 

Successfully set memory limits on the container at 600m for normal usage. Tested with a multi-profile preservation/archive format and works well. In Docker Stats, Linkwarden container hits 600m and drops to around 380/410m. Memory limit lower than that hinders workers. 500m is suboptimal, and 400m isn't functional. 

Memory limit set the Docker Compose:

```
deploy:
      resources:
        limits:
          memory: 600m
```

seals187:
ok nice, 
so a limit to 600m is stable ?

im actually trying linkwarden via pikapods, and I have there a pod (container) with 2 CPU and 2 GB of RAM, 
and the pod has only linkwarden on it, nothin else, is hitting 90% or almost 100% of RAM constantly... =)

psarchives:
> ok nice, so a limit to 600m is stable ?
> 
> im actually trying linkwarden via pikapods, and I have there a pod (container) with 2 CPU and 2 GB of RAM, and the pod has only linkwarden on it, nothin else, is hitting 90% or almost 100% of RAM constantly... =)

Yep stable on my server - I've been using it with that limit for about a month and its running perfect for me.

seals187:
ok cool, nice to now, will limit it the same way if I will set it up on my vps. thx =)

Issue #1281, opened 2025-07-04
RAM usage grows over time, but does not shrink


akramer-zibra:
This may not be a bug, but either an optimization. But for long running services, this issue may lead to service failure.

**Describe the bug**
The RAM usage of the linkwarden service grows over time, but does not seem to shrink. This may lead to disproportional RAM usage with longrunning services and will cost RAM, where it should not necessarily. Running this service in long term this also may lead to an out-of-memory situation. 

**To Reproduce**
Run linkwarden `v2.10.2` with `docker-compose.yml` (linkwarden + meilisearch + postgres) on any server.

```bash
NAME                       IMAGE                                   COMMAND                  SERVICE       CREATED       STATUS                 PORTS
linkwarden                 ghcr.io/linkwarden/linkwarden:v2.10.2   "docker-entrypoint.s…"   linkwarden    10 days ago   Up 10 days (healthy)   3000/tcp
linkwarden-meilisearch-1   getmeili/meilisearch:v1.12.8            "tini -- /bin/sh -c …"   meilisearch   10 days ago   Up 10 days             
linkwarden-postgres-1      postgres:16-alpine                      "docker-entrypoint.s…"   postgres      10 days ago   Up 10 days             
```

Running linkwarden service for 10 days uptime, linkwarden service consumes `1.289GiB` memory. 

```bash
CONTAINER ID   NAME                       CPU %     MEM USAGE / LIMIT   MEM %     NET I/O          BLOCK I/O         PIDS
0d769d1ab1c6   linkwarden                 0.46%     1.289GiB / 8GiB     16.12%    6.59GB / 457MB   1.07GB / 1.15GB   212
2f28bd7341f7   linkwarden-meilisearch-1   0.32%     786.4MiB / 8GiB     9.60%     1.37MB / 840kB   1.41GB / 3.73GB   42
05d6abc97eb9   linkwarden-postgres-1      0.00%     78.18MiB / 8GiB     0.95%     169MB / 6.01GB   137MB / 43.4MB    17
```

After restarting the whole service composition, linkwarden service runs with `560MiB` memory usage. This is ~ 56% less, which means RAM usage doubled within 10 days...

```
CONTAINER ID   NAME                       CPU %     MEM USAGE / LIMIT   MEM %     NET I/O           BLOCK I/O         PIDS
[...]

Issue #744, opened 2024-08-28
LinkWarden size and resource usage


oantonioo:
LinkWarden is huge, how is it possible that almost 6GB of space is needed to install this APP?

The CPU and RAM usage is also excessive.

Have you considered?
1 - Changing APACHE to NGNX.
2 - Giving the option to choose between Postgresql and **SQLite**, (and why not MariaDB?).
3 - Offering the possibility of not installing a database server and using one already installed on the network.

maliniak1:
I like the layout but resources consumption is a joke. 2vCPU and 4GB of RAM are not enough to run this.

thezuliano:
I would also add Mysql as a database option to choose, by other hand I don't think that the excessive usage of resources has to do with a database, generally Postgres and Mysql both have a good performance

daniel31x13:
We've cut it down to ~2GB in [v2.8.0](https://github.com/linkwarden/linkwarden/releases/tag/v2.8.0).

ash286:
I'd like to mention that Linkwarden (v2.9) is really really resource intensive especially on RPi4. I am running many other self-hosted containers but I haven't seen any other resource hungry containers like Linkwarden. Even when I am adding a link, I notice that it consumes quite a bit of CPUs (I am saving snapshot, HTML and PDF formats). Seach functionality is the most resource hungry and it max out CPU cores on RPi4.  I recently installed Linkwarden and really love it but I lookforward to optimised version of Linkwarden :)

RobertusIT:
Same here, 4Gb of ram are not enough and linkwarden don't respond. Is not possible to run with 4gb of ram or less? It too much r ram in my opinion, more than 4gb

mrkesu:
I was searching around to see what I was doing wrong because the resource usage really surprised me. Love the service, but I really hope this can be improved.

Are there any "fluff" options I can disable in the meantime to make it run more efficiently perhaps?

JamesMowery:
I just installed LinkWarden for the first time, and I was horrified with the resource usage (particularly RAM). It's like 5 - 10x any other self-hosted service I have installed (and I have around a dozen).

Unfortunately, I'm going to uninstall it, but I wanted to make sure that I wasn't experiencing a bug or something. The poor performance was confirmed on this thread.

Hopefully it is addressed at some point in the future and I'll give it another try then.

exentio:
[...]

PR #1665, opened 2026-04-14
Add on-demand browser lifecycle mode (BROWSER_LIFECYCLE=on-demand)

profiteroles1:
### Problem

Linkwarden keeps a headless Chromium process running permanently and restarts it every 30 minutes, even when no links are queued for processing. On memory-constrained hosts this consumes 300–800 MB of RAM at idle, and the 30-minute restart cycle causes sustained swap pressure on machines running other services alongside Linkwarden.

This is the single largest resource cost in Linkwarden and it's paid 24/7 regardless of actual usage. For self-hosted instances that save links sporadically (a few times a day or less), the browser sits idle >99% of the time.

Related: #744, #1431, #1396, #1281

### Solution

Add a new `BROWSER_LIFECYCLE` environment variable with two modes:

| Value | Behavior |
|---|---|
| `persistent` (default) | **Existing behavior, unchanged.** Browser launches at startup and stays alive with 30-minute rotation. Zero risk to current users. |
| `on-demand` | Browser launches only when unprocessed links are found. After the batch completes and no new links arrive within `BROWSER_IDLE_TIMEOUT_MS` (default: 60 seconds), the browser is closed. Relaunches automatically on the next save. |

On-demand mode still applies the same 30-minute max-age rotation to prevent memory leaks during long processing runs.

### Lifecycle diagram

```
                    ┌──────────────────────────────┐
                    │         No browser            │
                    │     (idle, ~50 MB total)      │
                    └──────────┬───────────────────┘
                               │
                         links found
                               │
                               ▼
                    ┌──────────────────────────────┐
                    │     Launch Chromium           │
                    │     Process batch             │
                    │     (~300-800 MB)             │
                    └──────────┬───────────────────┘
                               │
                     ┌─────────┴─────────┐
                     │                   │
                more links          no links for
                  queued          IDLE_TIMEOUT (60s)
                     │                   │
                     ▼                   ▼
                process next       Close browser
                   batch           → back to idle
```

### Configuration

```env
# Browser lifecycle mode: "persistent" (default, existing behavior) or "on-demand"
BROWSER_LIFECYCLE=on-demand

# How long to keep the browser alive after the last processed link before
# shutting it down. Prevents rapid launch/close cycles when saving multiple
# links in quick succession. Only applies in on-demand mode. Default: 60000 (1 minute).
BROWSER_IDLE_TIMEOUT_MS=60000
```

### Changes

- **`apps/worker/workers/linkProcessing.ts`** — Refactored into two functions: `linkProcessingPersistent()` (existing behavior, extracted as-is) and `linkProcessingOnDemand()` (new). The top-level `linkProcessing()` dispatches based on `BROWSER_LIFECYCLE`. No changes to the persistent code path.
- **`.env.sample`** — Added `BROWSER_LIFECYCLE` and `BROWSER_IDLE_TIMEOUT_MS` with documentation.

### What this does NOT change

- `browser.ts` — untouched
- `archiveHandler.ts` — untouched, still receives a `Browser` instance exactly as before
- Persistent mode — zero behavioral changes, remains the default
- No new dependencies

### Testing

1. **Persistent mode (regression):** Run without setting `BROWSER_LIFECYCLE` or with `BROWSER_LIFECYCLE=persistent`. Behavior is identical to current main — browser launches on startup, rotates every 30 minutes, processes links continuously.

2. **On-demand mode — idle:** Set `BROWSER_LIFECYCLE=on-demand`. Start the worker with no unprocessed links. Verify via `docker stats` that memory stays low (~50-150 MB) and no Chromium process is running. Logs should show normal polling with no browser launch messages.

3. **On-demand mode — link processing:** Save a link. Verify logs show "Launching browser on demand..." followed by normal processing. After processing completes and 60 seconds of idle, verify "Browser closed (on-demand idle shutdown)." appears and memory drops.

4. **On-demand mode — rapid saves:** Save 5 links in quick succession. Verify the browser launches once, processes all 5, then shuts down after the idle timeout — not launch/close per link.

5. **On-demand mode — 30-minute rotation:** Set `BROWSER_IDLE_TIMEOUT_MS=999999999` and queue enough links to keep the browser alive for 30+ minutes. Verify the rotation message still appears and the browser is relaunched.

6. **On-demand mode — browser crash:** Kill the Chromium process mid-archive. Verify the worker logs the disconnection and successfully relaunches on the next batch.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -18,6 +18,18 @@ DISABLE_NEW_SSO_USERS=
 MAX_LINKS_PER_USER=
 ARCHIVE_TAKE_COUNT=
 BROWSER_TIMEOUT=
+
+# Browser lifecycle mode. "persistent" (default) keeps Chromium running at all
+# times with a 30-minute restart rotation. "on-demand" launches Chromium only
+# when links need processing and shuts it down after an idle period, freeing
+# 300-800 MB of RAM on instances with sporadic usage.
+# BROWSER_LIFECYCLE=persistent
+
+# How long (ms) to keep the on-demand browser alive after the last processed
+# link. Prevents rapid launch/close cycles when saving multiple links in quick
+# succession. Only used when BROWSER_LIFECYCLE=on-demand. Default: 60000 (1 min).
+# BROWSER_IDLE_TIMEOUT_MS=60000
+
 IGNORE_URL_SIZE_LIMIT=
 NEXT_PUBLIC_DEMO=
 NEXT_PUBLIC_DEMO_USERNAME=
--- a/apps/worker/workers/linkProcessing.ts
+++ b/apps/worker/workers/linkProcessing.ts
@@ -4,17 +4,36 @@ import { delay } from "@linkwarden/lib/utils";
 import getLinkBatchFairly from "../lib/getLinkBatchFairly";
 import { launchBrowser } from "../lib/browser";
 import { countUnprocessedBillableLinks } from "../lib/countUnprocessedBillableLinks";
+import { Browser } from "playwright";
 
 const ARCHIVE_TAKE_COUNT = Number(process.env.ARCHIVE_TAKE_COUNT || "") || 5;
 const BROWSER_MAX_AGE_MS = 30 * 60 * 1000; // 30 minutes
 
+// On-demand mode: launch browser only when links need processing, close when
+// idle. Dramatically reduces memory usage on instances with sporadic traffic.
+// Default ("persistent") preserves existing always-on behavior.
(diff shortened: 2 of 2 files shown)


daniel31x13:
Hey good catch! I'll be merging this soon. Thanks.

daniel31x13:
@profiteroles1 Thanks for the PR! I tested this, but idle memory only changed from ~680 MB to ~640 MB with `BROWSER_LIFECYCLE=on-demand`, so the impact seems much smaller than described.

Issue #1704, opened 2026-05-31
Firefox-based stealth archiver backend (interest check)


feder-cr:
Firefox-based stealth archiver backend (interest check). draft PR #1705 has the proposal doc. addresses #1237 (Cloudflare/Turnstile capture failures); FlareSolverr #1251 only handles IUAM. optional, env-gated, no change to defaults. if not in scope i will close without noise.

Issue #1237, opened 2025-06-25
saving sites behind cloudflare challange and untrusted certs


lidoror:
How can I configure Linkwarden to properly save pages that are protected by Cloudflare challenges and also handle cases where the target site has an untrusted SSL certificate?

When I try to save pages from StackOverflow, I get the CloudFlare challenge instead of the real site.
When I try to save articles from sites with expired certificates, it saves nothing.
Is there a solution?

molnarti:
i think a solution could be if the the browser addon uploaded the page from the browsers local cache to Linkwarden, therefore the LW server would not need to download it itself (and bump into the Cloudflare challange or expired cert warning). Of course that would cover only the usecases when adding a link from the addon and not directly to LW, but its something that could be implemented relatively easily. 
Or you could use the already existing "Upload image" feature, but that only gives you an image without searchability.

molnarti:
i think this is resolved in 2.12 with the SingleFile integration

Bl4ckspell7:
Same issue when adding links directly in Linkwarden: CloudFlare challenge gets saved.

Would it be possible to add [FlareSolverr](https://github.com/FlareSolverr/FlareSolverr) integration? (option to add FlareSolverr API URL in Linkwarden settings)

gymzombie:
It looks like this has been done in a fork, but doesn't appear current: https://github.com/SteveImmanuel/linkwarden-flaresolverr

PR #1705, opened 2026-05-31
docs(proposal): Firefox-based stealth archiver backend

feder-cr:
opening as draft to check interest before building the env-gated branch.

would an optional Firefox-based stealth backend for the archiver be in scope, parallel to the current Chromium path in `apps/worker/lib/browser.ts`? selected via `ARCHIVE_BROWSER=invisible_firefox`, no change to defaults.

motivation: #1237. the headless Chromium archiver trips Cloudflare / Turnstile and captures the interstitial or a blank page instead of real content. the FlareSolverr route (#1251) only handles IUAM, not Turnstile / fingerprint checks. a patched Firefox 150 (feder-cr/invisible_firefox, MPL-2.0) with C++-source-level fingerprint patches avoids the JS-shim detection surface, so the page renders and the archive captures real content. returned page is a standard Playwright page, so screenshot / PDF / single-file / readability are unchanged.

this PR only adds a proposal doc so the idea has somewhere concrete to land. tracking issue: #1704

honest caveat: linkwarden is Node/TS, the wrapper is Python, so the wrapper won't drop in directly; the firefox binary is language-agnostic (executablePath + firefoxUserPrefs from playwright-node). if not in scope i'll close without noise.

### DIFF
--- a/INVISIBLE_FIREFOX_PROPOSAL.md
+++ b/INVISIBLE_FIREFOX_PROPOSAL.md
@@ -1,0 +1,29 @@
+# Firefox-based stealth archiver backend (proposal)
+
+> Status: Draft proposal
+> Created: 2026-05-31
+> Tracking issue: TBD
+
+## Goal
+
+Optional Firefox-based stealth backend for the archiver in `apps/worker/lib/browser.ts`, parallel to the current Chromium path. Selected via env, no change to defaults.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


daniel31x13:
Heya! Sorry for the delay. The problem with this approach is that it would require too much ongoing maintenance to keep the browser up to date. There’s a constant cat-and-mouse game between Cloudflare and web scrapers.

I was thinking more along the lines of using something like [CloakBrowser](https://github.com/CloakHQ/CloakBrowser)⁠￼.

feder-cr:
@daniel31x13 my project is similar to cloackbrowser but mine is open source

daniel31x13:
Cool! I'll check it out.

Btw do you have Discord or another way we could chat?

feder-cr:
@daniel31x13 telegram: @edefeder

Issue #1695, opened 2026-05-20
Inconsistency with jpg and jpeg file extensions


wildsheepz:
**Describe the bug**
Files are created as *.jpeg but deletion is *.jpg

**To Reproduce**
1. Add a link
2. See error in application logs

**Expected behavior**
No errors.

**Screenshots**
<img width="641" height="225" alt="Image" src="https://github.com/user-attachments/assets/6d6f8639-ea03-437f-a90a-e04d3d165b81" />

<img width="2367" height="120" alt="Image" src="https://github.com/user-attachments/assets/dc0e1a5f-5ef3-4156-b21e-d30cbbd113bd" />

**Desktop (please complete the following information):**
 - Version: Linkwarden v2.14.1

PR #1713, opened 2026-06-06
fix: ignore missing sibling archive files during cleanup

Anthomos:
## Summary
- ignore expected ENOENT errors during local archive cleanup
- keep logging real unlink failures
- add a focused regression test for missing sibling image files

## Why
Link cleanup already tries both .jpeg and .jpg variants. When one sibling file does not exist, local cleanup was logging a misleading error even though the situation is expected.

## Verification
- corepack yarn test packages/filesystem/removeFile.test.ts --run


Anthomos:
Closing this as superseded by #1725, which keeps the same ENOENT cleanup behavior for #1695 but has cleaner focused test setup. Keeping one PR open should make review easier.

PR #1715, opened 2026-06-08
fix: fall back when clipboard API is unavailable

### LINKED ISSUE #1707: Copy Icon Doesn't Work for Access Tokens
**Describe the bug**
When creating an access token a modal appears with the code and a copy icon. Clicking the copy icon doesn't copy the code to the clipboard. Not sure if this is just the self hosted version.

**To Reproduce**
1. Create Access Token
2. Click Icon, check clipboard.

**Expected behavior**
I would expect the code to be on the clipboard

 - Win 11, 64Bit, Portable Chrome Browser

Not a big deal, but wanted to report it. :)

Anthomos:
## Summary
- add a shared clipboard helper with an execCommand fallback
- use the fallback in CopyButton so token copy still works when the Clipboard API is blocked
- add a focused regression test covering the fallback paths

## Why
Fixes #1707. Some self-hosted or non-secure browser contexts block `navigator.clipboard.writeText()`, which prevented the access-token copy button from doing anything.

## Verification
- `corepack yarn test apps/web/lib/copyToClipboard.test.ts --run`

### DIFF
--- a/apps/web/components/CopyButton.tsx
+++ b/apps/web/components/CopyButton.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from "react";
 import { Button } from "./ui/button";
+import { copyTextToClipboard } from "@/lib/copyToClipboard";
 
 type Props = {
   text: string;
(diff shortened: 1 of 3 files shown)


Anthomos:
Reopened this change on top of `dev` to satisfy the branch policy and avoid the unrelated history in this branch. Replacement PR: #1718

Anthomos:
Closing in favor of #1718, which contains the same fix rebased onto `dev` and passes the branch-target rule.

PR #1720, opened 2026-06-10
fix(worker): add per-feed timeout to RSS polling to prevent loop freeze

### LINKED ISSUE #1721: RSS polling loop freezes permanently when a single feed hangs (no fetch timeout)
**Describe the bug**

The background RSS polling loop (`apps/worker/workers/rssPolling.ts`) stops permanently if a single subscribed feed's server accepts the connection but never sends a response.

`startRSSPolling()` fetches every subscription together in one `Promise.all` and only advances to the next cycle (`await delay(...)`) once that whole batch settles:

```ts
await Promise.all(
  rssSubscriptions.map(async (rssSubscription) => {
    const xml = await safeFetch(rssSubscription.url).then((res) => res.text());
    ...
  })
);
await delay(pollingIntervalInSeconds);
```

`safeFetch` uses **node-fetch v2, which has no default timeout**. If one feed's server accepts the TCP connection but never responds, that fetch stays pending forever, so the `Promise.all` never settles, `await delay(...)` is never reached, and the loop is frozen. No further cycles run and no new links are ingested until the worker is restarted. The existing per-feed `try/catch` does not help — it only catches rejections, and a hung request never rejects.

**To Reproduce**
1. Configure several RSS subscriptions.
2. Ensure one feed's endpoint accepts connections but stalls without responding (happens intermittently with some news RSS endpoints).
3. Wait for a polling cycle to hit that feed.
4. Observe that RSS polling stops entirely — no further `Processing new RSS feed items` logs, no new links — until the worker container is restarted.

**Expected behavior**
A single unresponsive feed should time out and be skipped. Polling should continue for all other feeds and keep cycling on schedule.

**Environment**
- Linkwarden: v2.14.x (self-hosted Docker); root cause also present on current `dev`
- Worker runtime: Node 20, node-fetch v2 (no default timeout)

**Additional context**
Observed in production: stalled feeds (some regional news RSS endpoints) froze polling for ~6 hours; a single container restart "fixed" it until the next stall.

Proposed fix in #1720 — wrap each feed fetch in an `AbortController`-based timeout (configurable via `RSS_FEED_TIMEOUT_SECONDS`, default 30s) and switch `Promise.all` to `Promise.allSettled`.

martiGIT:
## Problem

`startRSSPolling()` fetches every subscription together with `Promise.all` and only advances to the next cycle (`await delay(...)`) once that whole batch settles:

```ts
await Promise.all(
  rssSubscriptions.map(async (rssSubscription) => {
    const xml = await safeFetch(rssSubscription.url).then((res) => res.text());
    ...
  })
);
await delay(pollingIntervalInSeconds);
```

`safeFetch` uses **node-fetch v2, which has no default timeout**. If a single subscribed feed's server accepts the TCP connection but never sends a response, that one fetch stays pending forever. Because the whole batch is one `Promise.all`, the batch never settles — so `await delay(...)` is never reached and **the polling loop freezes indefinitely**. No further cycles run and no new links are ingested until the worker is restarted.

The existing per-feed `try/catch` does not help: it only catches *rejections*, and a hung request never rejects.

This is easy to hit in the wild — some news RSS endpoints intermittently accept connections and stall. In my deployment this froze RSS polling for ~6 hours (a single restart "fixed" it until the next stall).

## Fix

- Wrap each feed fetch in an `AbortController`-based timeout (default **30s**, configurable via `RSS_FEED_TIMEOUT_SECONDS`), cleared in `finally`.
- Switch `Promise.all` → `Promise.allSettled` so one failing feed can never reject the batch.

A stalled feed now aborts after the timeout, its error is logged by the existing `catch`, and the loop keeps cycling on schedule.

## Verification

Running this in production: stalled feeds (e.g. some regional news endpoints) now emit `AbortError: The user aborted a request.` and are skipped, while the polling loop continues firing every cycle on schedule — confirmed over a full night of 30-minute cycles with zero freezes (previously a single stall would freeze it until restart).

## Notes
- Minimal, behavior-preserving for healthy feeds.
- New env var `RSS_FEED_TIMEOUT_SECONDS` (optional, default 30) mirrors the existing `NEXT_PUBLIC_RSS_POLLING_INTERVAL_MINUTES` pattern.

Fixes #1721

### DIFF
--- a/apps/worker/workers/rssPolling.ts
+++ b/apps/worker/workers/rssPolling.ts
@@ -13,8 +13,7 @@ const pollingIntervalInSeconds =
 // its fetch pending forever. Because every feed is awaited together in one
 // Promise.all below, that one hung request makes the whole batch never settle,
 // which freezes the polling loop indefinitely (no further cycles ever run).
-const feedTimeoutInMs =
-  (Number(process.env.RSS_FEED_TIMEOUT_SECONDS) || 30) * 1000;
+const feedTimeoutInMs = 30_000;
 
 export async function startRSSPolling() {
   console.log("\x1b[34m%s\x1b[0m", "Starting RSS polling...");
--- a/apps/worker/workers/rssPolling.ts
+++ b/apps/worker/workers/rssPolling.ts
@@ -8,6 +8,14 @@ import { safeFetch } from "@linkwarden/lib/safeFetch";
 const pollingIntervalInSeconds =
   (Number(process.env.NEXT_PUBLIC_RSS_POLLING_INTERVAL_MINUTES) || 60) * 60; // Default to one hour if not set
 
+// Per-feed network timeout. safeFetch (node-fetch v2) has no default timeout,
+// so a single feed whose server accepts the connection but never responds keeps
+// its fetch pending forever. Because every feed is awaited together in one
+// Promise.all below, that one hung request makes the whole batch never settle,
+// which freezes the polling loop indefinitely (no further cycles ever run).
+const feedTimeoutInMs =
+  (Number(process.env.RSS_FEED_TIMEOUT_SECONDS) || 30) * 1000;
+
 export async function startRSSPolling() {
   console.log("\x1b[34m%s\x1b[0m", "Starting RSS polling...");
   while (true) {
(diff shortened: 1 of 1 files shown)


daniel31x13:
Thanks for pointing this out!

PR #1748, opened 2026-07-12
fix: cache validated DNS resolutions and reuse fetch agents in safeFetch

### LINKED ISSUE #1747: Archival of pages with many subresources times out: safeFetch performs 2+ uncached DNS lookups per request, which collapses on rate-limiting resolvers
**Describe the bug**

Links to pages with many subresources (GitHub repos, for example) consistently fail to preserve. All formats end up unavailable and the worker logs:

```
Failed Link: https://github.com/tw93/Mole
Reason: page.goto: Timeout 30000ms exceeded.
Call log:
  - navigating to "https://github.com/tw93/Mole", waiting until "domcontentloaded"
```

Simple pages preserve fine on the same instance.

**Root cause (traced and measured)**

Since v2.14, `protectPageRequests` re-fetches every page request through `safeFetch`. For each subresource that means at least two uncached DNS lookups: one in `assertUrlIsSafeForServerSideFetch` and one in the agent's custom `lookup` for every new socket (`createAgent` builds a fresh Agent per request, so sockets are never reused either). Archiving one GitHub page produces a burst of 50-80 DNS queries for a handful of hostnames.

Resolvers with rate limiting drop the excess. AdGuard Home, a common homelab DNS, ships with `ratelimit: 20` qps per client by default. Dropped UDP queries make getaddrinfo stall on OS retry timeouts, so request latencies quantize at ~4s steps. `domcontentloaded` waits for deferred scripts, and navigation blows past Playwright's default 30s timeout (there is no env knob for it; `BROWSER_TIMEOUT` only caps the whole archival at 5 minutes).

Measurements from inside the linkwarden container (v2.15.1):

| Test | Result |
|---|---|
| vanilla Playwright goto on github.com/tw93/Mole | 1.5s |
| the same goto through the interception | 21s, or timeout |
| `safeFetch` of 4 URLs, sequential | 0.2-1.1s each |
| `safeFetch` of the same 4 URLs, 24 in parallel | 8-12s each, latencies in exact 4s steps |
| 30 parallel `dns.lookup`, resolver = AdGuard (ratelimit 20) | median 4s, max 12s |
| 30 parallel `dns.lookup`, resolver = 1.1.1.1 directly | 68ms total |

**Fix (validated on a live v2.15.1 instance)**

Two changes, both keep the SSRF guarantees:

1. `packages/lib/ssrf.ts`: cache successful validated resolutions for a short TTL (30s) and deduplicate concurrent lookups for the same hostname (cache the in-flight promise, drop it on rejection; custom lookups used by tests bypass the cache). Serving a pinned, already-validated result during the TTL cannot be affected by DNS rebinding, since no re-resolution happens while it is served, which is in line with the intent of the rebinding fix from June.
2. `packages/lib/safeFetch.ts`: cache the http/https/proxy agents at module level with `keepAlive: true, maxSockets: 8` instead of creating a new Agent per request (and per redirect hop).

With both patches applied to a live v2.15.1 instance, and the rate-limited resolver left untouched, the same GitHub link preserves completely: screenshot, PDF, monolith, readable, and full text content.

I have the patches ready and can submit a PR.

**To Reproduce**

1. Point the host DNS at a resolver with rate limiting (AdGuard Home defaults work: `ratelimit: 20`).
2. Add a link to any sizable GitHub repo page.
3. Worker fails the link with `page.goto: Timeout 30000ms exceeded` after 30s.

**Expected behavior**

The page preserves, as it does when the interception layer is bypassed.

**Related issues**

- #1681 looks like the same failure seen from the user side (reported after v2.14, no diagnosis).
- #1721 / #1720 fixed the missing-timeout half of this problem for RSS polling; the archival path has the same issue plus the DNS burst described here.
- Older preservation-timeout reports (#1037, #655) predate the interception and likely have different causes.

- Server: Linkwarden v2.15.1 (ghcr.io, docker-compose: postgres:16-alpine, meilisearch v1.12.8), Debian 13 LXC on Proxmox, 4 cores / 6GB RAM
- Client: n/a, reproduces via API

Happy to provide the measurement scripts or full logs.

STuxuu-Legal:
Fixes #1747

Archiving pages with many subresources fails with `page.goto: Timeout 30000ms exceeded` because `safeFetch` performs at least two uncached DNS lookups per intercepted request (one in `assertUrlIsSafeForServerSideFetch`, one in the agent lookup for every new socket) and creates a new Agent per request and per redirect hop, so connections are never reused. On rate-limiting resolvers (AdGuard Home ships with `ratelimit: 20` by default) the burst gets dropped and getaddrinfo stalls on multi-second retry timeouts. Full analysis and measurements are in the issue.

**Changes**

- `packages/lib/ssrf.ts`: successful resolutions through the default lookup are cached for 30s, and concurrent lookups for the same hostname share one in-flight promise. Failed resolutions are not cached. Custom lookups (used by the tests) bypass the cache. Rebinding protection stays intact: during the TTL a pinned, already-validated result is served and no re-resolution happens.
- `packages/lib/safeFetch.ts`: http, https and proxy agents are cached at module level with keep-alive and a per-host socket cap, instead of being created per request.

**Validation** (live v2.15.1 instance, docker, resolver rate limit left in place)

- before: GitHub repo pages consistently fail with the navigation timeout; 24 parallel safeFetch calls take 8-12s each, with latencies quantized in 4s steps
- after: the same page preserves completely (screenshot, PDF, monolith, readable and text content)

Happy to adjust the TTL, cache size or socket cap if you prefer different values.

### DIFF
--- a/packages/lib/safeFetch.ts
+++ b/packages/lib/safeFetch.ts
@@ -56,36 +56,62 @@ function createSafeLookup() {
   return lookup;
 }
 
+// Agents are cached at module level so connections are reused across
+// requests (keep-alive). Creating a new Agent per request forces a fresh
+// TCP+TLS handshake for every fetch, which on pages with dozens of
+// subresources adds 15-25s and pushes page.goto past its navigation
+// timeout. Connection reuse also cannot be affected by DNS rebinding,
+// since no re-resolution happens on a reused socket.
+let cachedProxyAgent: HttpsProxyAgent<string> | SocksProxyAgent | undefined;
+let cachedHttpAgent: http.Agent | undefined;
+let cachedHttpsAgent: https.Agent | undefined;
+
 function createAgent(target: URL) {
   if (process.env.PROXY) {
-    const proxy = new URL(process.env.PROXY);
+    if (!cachedProxyAgent) {
+      const proxy = new URL(process.env.PROXY);
 
-    if (process.env.PROXY_USERNAME) {
-      proxy.username = process.env.PROXY_USERNAME;
-      proxy.password = process.env.PROXY_PASSWORD || "";
+      if (process.env.PROXY_USERNAME) {
+        proxy.username = process.env.PROXY_USERNAME;
+        proxy.password = process.env.PROXY_PASSWORD || "";
+      }
+
+      const ProxyAgent = proxy.protocol.includes("http")
+        ? HttpsProxyAgent
+        : SocksProxyAgent;
+
+      cachedProxyAgent = new ProxyAgent(proxy.toString());
     }
 
-    const ProxyAgent = proxy.protocol.includes("http")
-      ? HttpsProxyAgent
-      : SocksProxyAgent;
-
-    return new ProxyAgent(proxy.toString());
+    return cachedProxyAgent;
   }
-
-  const lookup = createSafeLookup();
 
   if (target.protocol === "http:") {
-    return new http.Agent({ lookup });
+    if (!cachedHttpAgent) {
+      cachedHttpAgent = new http.Agent({
+        lookup: createSafeLookup(),
+        keepAlive: true,
+        maxSockets: 8,
+      });
+    }
+
+    return cachedHttpAgent;
   }
 
-  return new https.Agent({
-    lookup,
-    rejectUnauthorized:
-      process.env.ALLOW_INSECURE_TLS === "true" ||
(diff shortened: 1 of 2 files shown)


LukaHummel:
@STuxuu-Legal Are you going to continue with the PR? I'm encountering a similar error (see #1778) and might check out your PR to see if it fixes it or submit my own.

